PR #18, opened 2020-09-20
Fix palette (top-right) close buttons

### LINKED ISSUE #11: Palette close buttons don't work
The close buttons on the Tray*, Markers, and Messages palettes don't work.  If I click them, the palette closes for an instant, and then reappears.

wsu-cb:
Fixes #11


DLLarson (commented):
Looks good.

PR #22, opened 2020-09-26
avoid gap between CPlayBoardView and side palettes

### LINKED ISSUE #20: width of right side palettes is not restored by CWinStateManager
Is the following a bug or a feature?

1)  Open a .gam
2)  Make the right side dock "very wide"
3)  Save the .gam
4)  Make the right side dock "normal width"
5)  Close the .gam
6)  Open the .gam
-->
The right side dock is "normal width".  The CWinStateManager has not restored the palette layout as it was when the .gam was saved.

Note:  If this is a feature, then there is a bug in the feature:  If a CPlayBoardView is the topmost child frame when the .gam is saved, its width at the time of the save is restored when the .gam is opened.  This results in a weird display because the board is narrow, and the right dock is narrow, leaving a gap between them where the project view shows through.  This can be fixed by adding
`(DYNAMIC_DOWNCAST(CMDIFrameWndEx, AfxGetMainWnd()))->RecalcLayout();`
to the end of CWinStateManager::RestoreStateOfDocumentFrames().

wsu-cb:
Fixes #20

### DIFF
--- a/GShr/WinState.cpp
+++ b/GShr/WinState.cpp
@@ -136,6 +136,7 @@ void CWinStateManager::RestoreStateOfDocumentFrames()
             }
         }
     }
+    (DYNAMIC_DOWNCAST(CMDIFrameWndEx, AfxGetMainWnd()))->RecalcLayout();
 }
 
 ///////////////////////////////////////////////////////////////////////////


DLLarson (approved):
I verified the fix against the issue's description to reproduce the gapping. The gapping no longer occurs.

-Dale

PR #24, opened 2020-10-05
force a redraw to work around issue #23

### LINKED ISSUE #23: window redraw problem
CBDesign can have a window redraw problem as seen here:  image

To reproduce this:

1)  In CBDesign, open the .gbx in [unnamed1.zip](https://github.com/CyberBoardPBEM/cbwindows/files/5315884/unnamed1.zip)
2)  Open board001
3)  Use the mouse to move board001 into a separate tab group (as in image)
4)  Open board002

This will result in the redraw problem shown above.

Note that this does not seem to happen in CBPlay.  If I make a scenario based on the same .gbx, and open the boards in the same way, the display looks normal.  I haven't found any reason for the difference.

wsu-cb:
I still can't explain why CBDesign behaves this way, so here's a kludge to get around the problem.

Fixes #23


DLLarson (approved):
Tested. Works as specified.

-Dale

PR #25, opened 2020-10-10
Fix mdi vestiges

wsu-cb:
Here's a suggestion for how to update the menus to reflect the change from MDI to tabbed windows.

Please note that this is a draft because 
1)  branch master currently has another pull request pending, so this should be rebased after that pull request is merged. 
2)  I don't know if we actually want the menu to enable splitting the tab groups
3)  I don't know if we want to use new menu ID values rather than reusing the old ones

### DIFF
--- a/GM/Cbdesign.rc
+++ b/GM/Cbdesign.rc
@@ -241,10 +241,8 @@ BEGIN
     END
     POPUP "&Window"
     BEGIN
-        MENUITEM "&Cascade",                    ID_WINDOW_CASCADE
-        MENUITEM "&Tile Horizontally",          ID_WINDOW_TILE_HORZ
-        MENUITEM "&Tile Vertically",            ID_WINDOW_TILE_VERT
-        MENUITEM "&Arrange Icons",              ID_WINDOW_ARRANGE
+        MENUITEM "Split Tabs &Horizontally",    ID_WINDOW_TILE_HORZ
+        MENUITEM "Split Tabs &Vertically",      ID_WINDOW_TILE_VERT
         MENUITEM SEPARATOR
         MENUITEM "T&ool Palette",               ID_WINDOW_TOOLPAL
         MENUITEM "Co&lor Palette",              ID_WINDOW_COLORPAL
@@ -1259,10 +1257,8 @@ END
 STRINGTABLE
 BEGIN
     ID_WINDOW_NEW           "Open another window for the active document"
-    ID_WINDOW_ARRANGE       "Arrange icons at the bottom of the window"
-    ID_WINDOW_CASCADE       "Arrange windows so they overlap"
-    ID_WINDOW_TILE_HORZ     "Arrange windows as non-overlapping tiles"
-    ID_WINDOW_TILE_VERT     "Arrange windows as non-overlapping tiles"
+    ID_WINDOW_TILE_HORZ     "Split active tab into a new group horizontally"
+    ID_WINDOW_TILE_VERT     "Split active tab into a new group vertically"
     ID_WINDOW_SPLIT         "Split the active window into panes"
 END
 
--- a/GM/FrmMain.cpp
+++ b/GM/FrmMain.cpp
@@ -44,9 +44,9 @@ static char THIS_FILE[] = __FILE__;
 ///////////////////////////////////////////////////////////////////////
 // CMainFrame
 
-IMPLEMENT_DYNAMIC(CMainFrame, CMDIFrameWndEx)
+IMPLEMENT_DYNAMIC(CMainFrame, CMDIFrameWndExCb)
 
-BEGIN_MESSAGE_MAP(CMainFrame, CMDIFrameWndEx)
+BEGIN_MESSAGE_MAP(CMainFrame, CMDIFrameWndExCb)
     ON_WM_CREATE()
     ON_WM_CLOSE()
     ON_WM_PALETTECHANGED()
--- a/GM/FrmMain.cpp
+++ b/GM/FrmMain.cpp
@@ -107,7 +107,6 @@ static CRuntimeClass *tblBrd[] = { RUNTIME_CLASS(CBrdEditView), NULL };
 
 CMainFrame::CMainFrame()
 {
-    DockInit();
     m_bColorPalOn = TRUE;
     m_bTilePalOn = TRUE;
 }
--- a/GM/FrmMain.h
+++ b/GM/FrmMain.h
@@ -24,6 +24,8 @@
 #ifndef _FRMMAIN_H
 #define _FRMMAIN_H
 
+#include "LibMfc.h"
+
 #ifndef     _GMDOC_H
 #include    "gmdoc.h"
 #endif
@@ -40,7 +42,7 @@
 #include    "frmdocktile.h"
 #endif
 
-class CMainFrame : public CMDIFrameWndEx
+class CMainFrame : public CMDIFrameWndExCb
 {
     DECLARE_DYNAMIC(CMainFrame)
 public:
--- a/GP/Cbplay.rc
+++ b/GP/Cbplay.rc
@@ -245,10 +245,8 @@ BEGIN
     END
     POPUP "&Window"
     BEGIN
-        MENUITEM "&Cascade",                    ID_WINDOW_CASCADE
-        MENUITEM "Tile &Horizontally",          ID_WINDOW_TILE_HORZ
-        MENUITEM "Tile &Vertically",            ID_WINDOW_TILE_VERT
-        MENUITEM "&Arrange Icons",              ID_WINDOW_ARRANGE
+        MENUITEM "Split Tabs &Horizontally",    ID_WINDOW_TILE_HORZ
+        MENUITEM "Split Tabs &Vertically",      ID_WINDOW_TILE_VERT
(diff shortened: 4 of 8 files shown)


DLLarson:
>so this should be rebased after that pull request is merged.

I totally missed that pull request! I'll deal with it by mid-day tomorrow. Sorry about that!

-Dale

DLLarson:
This will need to be rebased against master which now contains https://github.com/CyberBoardPBEM/cbwindows/pull/24.


-Dale

DLLarson:
How does the MDI behavior line up with the currently released CB built on MFC-CodeJock?

-Dale

wsu-cb:
> How does the MDI behavior line up with the currently released CB built on MFC-CodeJock?
> 

The CodeJock version is an odd mix of "real" MDI and a tabbed interface.  The MFC-only version is a pure tabbed interface.  In both, there aren't any real icons as in "real" MDI.  Since CodeJock has actual MDI-style frames around the child windows, it supports minimize/maximize/restore window operations, but MFC-only does not.  Likewise, CodeJock supports "Cascade", MFC-only does not. 

MFC-only master does nothing with the Tile Horizontal/Vertical menu commands.  My MFC-only mdi branch replaces them with "Split" commands that move the current tab into a separate tab group.  To be honest, I would rather see something closer to "real" MDI/CodeJock, which has Tile commands that moved every window into a separate tab group, but that's more work since I don't see that the MFC has any explicit support for evenly dividing the the available space among tab groups.  I wasn't ambitious enough to do the tile layout work myself, so I just exposed the provided `CMDIFrameWndEx::MDITabNewGroup()` through the menus.

Should we just remove the MDI menu commands rather than take this half-baked step?  As I said in https://github.com/CyberBoardPBEM/cbwindows/pull/25#issue-501034101, I'm not sure if it's worth exposing `CMDIFrameWndEx::MDITabNewGroup()` through the menus.  The user can already split the tabs using the mouse in branch master.  On the other hand, that's not obvious to the user.

DLLarson:
I'm conflicted about this. I just went through the MFC source code and there is definitely support code for these particular messages. I set a breakpoint in the code in `winmdi.cpp` line 1101:

```
BOOL CMDIFrameWnd::OnMDIWindowCmd(UINT nID)
{
	ASSERT(m_hWndMDIClient != NULL);  // Line 1101

	UINT msg;
	UINT wParam = 0;
	switch (nID)
	{
	default:
		return FALSE;       // not for us
	case ID_WINDOW_ARRANGE:
		msg = WM_MDIICONARRANGE;
		break;
	case ID_WINDOW_CASCADE:   // <== THIS IS THE MENU COMMAND I SELECTED
		msg = WM_MDICASCADE;
		break;
	case ID_WINDOW_TILE_HORZ:
		wParam = MDITILE_HORIZONTAL;
		// fall through
	case ID_WINDOW_TILE_VERT:
		ASSERT(MDITILE_VERTICAL == 0);
		msg = WM_MDITILE;
		break;
	}

	::SendMessage(m_hWndMDIClient, msg, wParam, 0);
	return TRUE;
}
```
And it steps through code to ultimately send a MDI message to the Windows internal handler for MDI windows.

The mystery to me is why that code isn't processing the request.

Something's amiss...

-Dale

wsu-cb:
That's what I found as well. I haven't used MFC much since 20 years ago, so I haven't paid much attention to its evolution.  I just assumed that `CMDIFrameWndEx` with tabs isn't a true MDI window, and that's why the MDI messages failed.

If you are interested in this bug, feel free to reject this draft pull request, and re-assign the bug to yourself.  I only put this forward in an attempt to get rid of obviously incorrect behavior so we could get the new CB to a releasable state.

wsu-cb:
Does the new MFC have any sample programs using the tabbed MDI interface?  Or wizards for generating programs using it?  If so, seeing how those programs behave might also give some hints.

DLLarson:
I have been messing around with the VS2019 MFC App wizard but it's quite different from the old days (more options and complication without any popup help for what the items do.) I haven't reproduced a classic MDI app yet (two attempts so far.)

-Dale

wsu-cb:
> I have been messing around with the VS2019 MFC App wizard but it's quite different from the old days (more options and complication without any popup help for what the items do.) I haven't reproduced a classic MDI app yet (two attempts so far.)
> 

But can it produce a TABBED MDI app?  And if so, how does that app behave?  If the wizard produces a tabbed MDI app that behaves well, then it doesn't seem like it matters whether it can produce a classic MDI app.  We would "just" have to figure out what CB was doing differently from the wizard-produced tabbed MDI app.

DLLarson:
I've managed to create an MDI application using the VS2019 MFC App wizard tool.
See attached.

[MFC-MDI-Application.zip](https://github.com/CyberBoardPBEM/cbwindows/files/5366804/MFC-MDI-Application.zip)

-Dale

wsu-cb:
> I've managed to create an MDI application using the VS2019 MFC App wizard tool.

That looks like a traditional MDI program.  Does that mean you are going to switch CB back to traditional MDI rather than use the tabbed interface currently in branch master?

DLLarson:
>That looks like a traditional MDI program. Does that mean you are going to switch CB back 
>to traditional MDI rather than use the tabbed interface currently in branch master?

For me the first goal was to get the open source version working in a similar fashion to the currently shipping version.  However, the enhanced capabilities may offer something more compelling if we gave up the pure MDI features such as tiling or cascade. 

So I'd like to see if we can even restore classic MDI operation without breaking the UI. After which we can see what we've lost in that process.

At this point I just don't have a clear answer for this other than "Do no harm.".

-Dale

DLLarson:
Hi Bill,

I've done some additional investigation with the MDI feature as supported by the latest MFC. First I created an old style MDI application using the app wizard. That worked fine as can be seen with the code I posted earlier above. Then I changed the MDI windows classes to add the `Ex` suffix that pulls in the new features. Unlike CodeJock, this seems to completed disable the Windows supported MDI functionality due to them adding the tab control into the MDI Client hosting window.

As we seen above, that turns off the MDI feature within the raw Windows API's. So if it's a given that we can't fix this easily we should probably look at exploiting the new MFC's capabilities to provide similar features--which may come down to just splitting the workspace. I don't see any way to provide non-tiled windows or icons.

Am I missing anything here?

-Dale

wsu-cb:
> Hi Bill,
> 
> As we seen above, that turns off the MDI feature within the raw Windows API's. So if it's a given that we can't fix this easily we should probably look at exploiting the new MFC's capabilities to provide similar features--which may come down to just splitting the workspace. I don't see any way to provide non-tiled windows or icons.
> 
> Am I missing anything here?
> 
> -Dale

This sounds like you are seeing much the same thing I did.

Have you looked at the system behavior resulting from applying this Pull Request?  How do you feel about these changes?

DLLarson:
Ok... I've looked at the PR and I like pretty much like it. The choices appear to be an XOR of either vertically or horizontally tiled. It doesn't look like I can vertically split a horizontally split thought to get classic tile style windowing. But if that's not possible I think that's fine. Once people get used to the new scheme, it is superior to the historical MDI approach. You just can't cascade or icon-ize the windows. They are simply always "docked". It's also consistent with the way toolbars and tool palettes work (other than those can also be floated in a new windows frame.

I like that I can do the split from the Window menu. That makes the capability explicit.

I'm good with moving forward with this if you're not uncomfortable with it.

-Dale

wsu-cb:
> Ok... I've looked at the PR and I like pretty much like it. The choices appear to be an XOR of either vertically or horizontally tiled. It doesn't look like I can vertically split a horizontally split thought to get classic tile style windowing. But if that's not possible I think that's fine. Once people get used to the new scheme, it is superior to the historical MDI approach. You just can't cascade or icon-ize the windows. They are simply always "docked". It's also consistent with the way toolbars and tool palettes work (other than those can also be floated in a new windows frame.
> 
> I like that I can do the split from the Window menu. That makes the capability explicit.
> 
> I'm good with moving forward with this if you're comfortable with it.
> 
> -Dale

Yes, as far as I can tell the new MFC doesn't support simultaneous horizontal/vertical tab groups.

However, I'm not quite ready to call this PR ready-for-review.

First, this needs rebasing.

Second, I have found and fixed one small bug.

Third, how do you feel about reusing the "standard" ID_WINDOW_TILE_HORZ/VERT ids for the "split" menu items?

DLLarson:
>Third, how do you feel about reusing the "standard" ID_WINDOW_TILE_HORZ/VERT ids for the "split" menu items?

I'm fine with it. There's no real reason to create something that means basically the same thing.

-Dale

wsu-cb:
Can you see why the GitHub build is failing?  The code builds successfully on my local machine.  Also, the report sugggests GitHub has changed their environment, but I have no experience with GitHub builds.

DLLarson:
It looks like GitHub changed their check build environment to address a security problem:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

I'll check it out and see if I can fix the actions stuff.

-Dale

DLLarson:
Ok... I think I fixed it.

You'll need to rebase against master again to pick up the fix.

-Dale

DLLarson (approved):
Looks good to me.

Works great!

-Dale

PR #27, opened 2020-10-17
be more explicit about nature of ObjectID

wsu-cb:
I made some changes to clarify how ObjectID works (or at least how I think it works).  Could you review this and see if this looks correct to you?  I would like to do similar work for other "generic integer" types (PieceID, MarkID, GameElement) as well.  This is preparation for moving the 16 bit identifiers to 32 bit since I hope this will make it easier to understand how the identifiers currently behave.

### DIFF
--- a/GShr/DrawObj.h
+++ b/GShr/DrawObj.h
@@ -42,6 +42,7 @@ class CSelection;
 #ifdef GPLAY
 class CPlayBoardView;
 class CGamDoc;
+class ObjectID;
 #else
 class CBrdEditView;
 #endif
@@ -98,7 +99,7 @@ public:
     virtual CRect GetEnclosingRect() { return m_rctExtent; }
     virtual BOOL HitTest(CPoint pt) { return FALSE; }
 #ifdef GPLAY
-    virtual DWORD GetObjectID() { return 0; }
+    virtual ObjectID GetObjectID();
     virtual void MoveObject(CPoint ptUpLeft);
 #endif
     virtual void OffsetObject(CPoint offset);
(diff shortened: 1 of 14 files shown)


GShr/DrawObj.h
```diff
@@ -162,6 +163,42 @@ class CDrawObj
 
 ///////////////////////////////////////////////////////////////////////
```
DLLarson, line 165: After our email discussions I think it would be good to summarize the reason for ObjectID's now that they are more explicit. Something about providing unique ID's in a manner that can be performed in a distributed manner that attempts to avoid ID collisions. At a higher level they are used to manage the maintenance of objects that pop into existence and then later can be destroyed in a distributed manner as well. Current and possible uses are creating markers, lines (used for plotting moves), grease pencil for freehand annotation. attaching disposible multimedia clips, etc...

DLLarson (changes requested):
I like the overall approach. More modern. I hope we'll be safe using C++11 features from here out. I think we are (after all it was nine years ago since adopted!)

I added one requested change having to do with commenting the new class.

-Dale

GShr/DrawObj.h
DLLarson: I ran into an operational issue. If I drag some markers onto a board. I can't delete them when selected.

Also, when I do a Ctrl drag of a marker onto the board it is suppose to prompt for a number of items to create but it asserts.

-Dale

wsu-cb: I just tried the POG test, and was able to delete markers.  Could you please double-check that you were actually trying to delete markers, not pieces?  I notice that the POG test has PIECES that are in a group named "Control Markers".

DLLarson: OMG!...

Egg on my face. I "bought" the idea that those pieces were markers because they said they were!

Sorry about that. My bad and all that.

Ummmm... Never mind...

-Dale

DLLarson (approved):
Looks good!

-Dale

wsu-cb:
I am planning to do more of this kind of work before tackling the actual conversion to 32bit ids.  Do I need to create a new branch for continuing on, or can I continue using this branch?  Assuming I can use this branch, can I continue to use this pull request, or do I need a new pull request now that this one has been closed?

DLLarson:
>Do I need to create a new branch for continuing on, or can I continue using this branch?

I think as a matter of good practice you should use a new branch with a descriptive name for any new work. At times you may have many so called "topic branches" in play. Best to retire the completed ones.

Note that this is not a rule. Certainly GIT doesn't require it.

-Dale

wsu-cb:
I had planned to work on clarifying the behavior of all the "simple integer type" identifiers, and similarly with things like CGrafixListBox::m_pItemMap, in this branch, but I can certainly use new branches for each identifier type.  However, I think the branches will have order-of-merge dependencies because a lot of files use more than one of these types.  To me, that suggests this should be a single branch rather than a bunch of branches.

DLLarson:
It's true that topic branches can go stale but that's what the `git rebase` command is for.

For me the ability to jump around in my work was essential (back in my working days that is) so this happened a lot. Some branches were long term projects, some where quick fixes, etc... 

Rebase straightens things out quite nicely. It's a very powerful command. `gitk --all` is handy to get a feel for the commits you are behind.

-Dale

PR #28, opened 2020-10-17
Don't let MFC display context menu when we perform background-color-select

### LINKED ISSUE #26: CBDesign color palette right click does two things
With the CBDesign color palette in the usual right hand dock, right clicking on a color to select the background color also brings up the popup menu to change the window docking state.

wsu-cb:
Fix #26


DLLarson (approved):
Looks good!

-Dale

Issue #19, opened 2020-09-20
UI Scaling Tiny on 4K (High DPI) screens


DLLarson:
CBDesign and CBPlay have some visual elements rendered very small on a 4K display.

![image](https://user-images.githubusercontent.com/175455/93713946-ec4bb380-fb24-11ea-877d-4def96f20501.png)

The workaround is to set application compatibility like so:

![image](https://user-images.githubusercontent.com/175455/93714026-5d8b6680-fb25-11ea-95fa-04cf18204017.png)

Unfortunately this makes the app "fuzzy":

![image](https://user-images.githubusercontent.com/175455/93714055-a04d3e80-fb25-11ea-9e0c-f00eaaf5cf50.png)

It would be best if the program could properly adapt to high DPI displays automatically.

DLLarson:
I've discovered that when using the high DPI compatibility workaround the XOR-ed tracking rects are far away from the proper screen coordinates. 

Another strong reason to make the program high DPI aware!

-Dale

DLLarson:
### A collection of links to Win10 high DPI operation

High DPI programming...

>MFC
Although we’ve added per-monitor DPI scaling support to Win32, WPF and WinForms, MFC does not natively support this functionality.

https://developercommunity.visualstudio.com/content/problem/692467/visual-studio-2019-1616-prof-high-dpi-problem-in-m.html

https://blogs.windows.com/windowsdeveloper/2016/10/24/high-dpi-scaling-improvements-for-desktop-applications-and-mixed-mode-dpi-scaling-in-the-windows-10-anniversary-update/

https://blogs.windows.com/windowsdeveloper/2017/05/19/improving-high-dpi-experience-gdi-based-desktop-apps/

https://blogs.windows.com/windowsdeveloper/2017/04/04/high-dpi-scaling-improvements-desktop-applications-windows-10-creators-update/

https://blog.softwareverify.com/how-to-make-your-mfc-or-non-mfc-program-support-high-dpi-monitors-the-easy-way/

User information...

https://www.windowscentral.com/how-change-high-dpi-settings-classic-apps-windows-10-april-2018-update

-Dale

dlazov:
Greetings I want to add to this problem as well. I am running a 4K monitor (27" inch).

This first screen shot is from the CBDesigner and the second screen shot is from CB 3.50  after open up the map you can see the differences between the two

CB 3.10
[...]

PR #30, opened 2020-10-24
improve type safety and clarity of drag-and-drop

wsu-cb:
Here is some work to improve the type-safety and clarity of drag-and-drop code.  However, I admit it's only clearer if one is comfortable with templates.

Also, I believe the comments in DragDrop.h describing the drag-specific info did not always match the running code.  I have updated the comments to reflect my understanding of the actual code.  If the comments represent actual desired behavior, the comments will need to be reverted to the previous version, and the code modified to behave as desired.

### DIFF
--- a/GM/LBoxTile.cpp
+++ b/GM/LBoxTile.cpp
@@ -140,15 +140,15 @@ BOOL CTileListBox::OnDragSetup(DragInfo* pDI)
     if (IsMultiSelect())
     {
         pDI->m_dragType = DRAG_TILELIST;
-        pDI->m_dwVal = (DWORD)GetMappedMultiSelectList();// TileID array
-        pDI->m_pObj = (void*)m_pDoc;
+        pDI->GetSubInfo<DRAG_TILELIST>().m_tileIDList = GetMappedMultiSelectList();// TileID array
+        pDI->GetSubInfo<DRAG_TILELIST>().m_gamDoc = m_pDoc;
         pDI->m_hcsrSuggest = g_res.hcrDragTile;
     }
     else
     {
         pDI->m_dragType = DRAG_TILE;
-        pDI->m_dwVal = (DWORD)GetCurMapItem();      // The TileID
-        pDI->m_pObj = (void*)m_pDoc;
+        pDI->GetSubInfo<DRAG_TILE>().m_tileID = GetCurMapItem();      // The TileID
+        pDI->GetSubInfo<DRAG_TILE>().m_gamDoc = m_pDoc;
         pDI->m_hcsrSuggest = g_res.hcrDragTile;
     }
     return TRUE;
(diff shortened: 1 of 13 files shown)


DLLarson (commented):
I'm having a variety of odd behaviors in CBDesign and CBPlay. When using CBDesign and the Path of Glory gamebox I get invalid argument errors when trying drag tile onto the map.

In CBPlay when I drag markers onto the map they seem to work but I can't find the marker. If I use Ctrl-Drag on a marker it prompts me (as it should) for a marker count to create but I can't find the markers on the map. They aren't dropped where I placed them.

-Dale

wsu-cb:
> I'm having a variety of odd behaviors in CBDesign and CBPlay. When using CBDesign and the Path of Glory gamebox I get invalid argument errors when trying drag tile onto the map.
> 

I can reproduce this.


> In CBPlay when I drag markers onto the map they seem to work but I can't find the marker. If I use Ctrl-Drag on a marker it prompts me (as it should) for a marker count to create but I can't find the markers on the map. They aren't dropped where I placed them.
> 

But I can't reproduce this.  Both single marker and multi-marker drops put their markers right where they should for me, even when I have scrolled the map away from the top-left corner, and even when I use the half-size and tiny boards.  Do you have any more information on how to reproduce this?

DLLarson:
>But I can't reproduce this.

Huh.... Well now I can't reproduce either!

I wonder if my running on a 4K screen caused that. Sometimes desktop applications can have weird scaling issues depending how the system wants to automatically adjust for High-DPI scaling.  There are definitely issues with that:

See https://github.com/CyberBoardPBEM/cbwindows/issues/19

Well, it's gone at the moment. If it occurs again I'll be more attentive to what's going on.

-Dale

-Dale

DLLarson (approved):
Ok. I'm not having any odd issues with behavior now.

I like the modernization of the code. As you say templates are somewhat black arts for folks but it's the type safety that's important here. The explicit callouts when invoking a template clarify what actually happening in the code.

With regard to the various ASSERT's about seemingly dead code paths I'll let that stand to see if we get triggers. Some of the code was built with a nod toward possible needs and the code may not be used at present. On the other hand the code could be logically dead through upstream coding.

Nice work!

-Dale

wsu-cb:
> Ok. I'm not having any odd issues with behavior now.
> 

Good!


> I like the modernization of the code. As you say templates are somewhat black arts for folks but it's the type safety that's important here. The explicit callouts when invoking a template clarify what actually happening in the code.
> 

I'm glad you like what I'm doing since I am working on more of this.

> With regard to the various ASSERT's about seemingly dead code paths I'll let that stand to see if we get triggers. Some of the code was built with a nod toward possible needs and the code may not be used at present. On the other hand the code could be logically dead through upstream coding.
> 
> Nice work!
> 

Thank you

Issue #31, opened 2020-11-12
possible CMarkerPalette::UpdatePaletteContents() logic error


wsu-cb:
The code for CMarkerPalette::UpdatePaletteContents() looks strange to me.  I think it is intending to remember the current selected marker group, update the palette content, and then re-select the previously remembered marker group.  However, it looks to me like the logic has a bug.  I think the current code actually re-selects the combo-box index which is numerically equal to the marker group index, rather than the combo-box index whose item data contains the marker group index.

In other words, I think CMarkerPalette::UpdatePaletteContents() should look slightly more like CTrayPalette::UpdatePaletteContents().

Or am I confused?

DLLarson:
>I think the current code actually re-selects the combo-box index which is numerically equal to the marker group index, rather than the combo-box index whose item data contains the marker group index. .... Or am I confused?

You are correct but it turns out that the combo index is equal to the marker index due to a call to `CMarkerPalette::LoadMarkerNameList()` method. It sets them to be identical.

However, that then begs the question: why store the group index at all? 

I believe I was planning for sorting the combo alphabetically or allowing users to set the order rather than showing them in GameBox created order. (The same is true for piece name combos and tile name combos.) It always annoyed me that the combo box order is fixed by the creation order. It's kind of a loose end in the implementation.

With that in mind--you're correct that this would then be a bug. If you'd like to correct the code I would appreciate it!

-Dale

DLLarson:
I need to correct this a bit:

>I believe I was planning for sorting the combo alphabetically or allowing users...

After stepping through the code changes I discovered that the combo boxes are already sorted alphabetically due to the `CBS_SORT` flag when the control was created in `OnCreate`. This means the group numbers stored in the combo items are not necessarily the same as the item's index in the box.

So that makes your original observation true that it's a definite bug. 

-Dale

DLLarson:
Code checked in. 

Closing issue since it wasn't closed automatically. PR's need to be associated with the issue for that to happen.

-Dale

PR #32, opened 2020-11-15
fix logic for maintaining selected marker group

wsu-cb:
Fix issue #31


DLLarson (approved):
Looks good!

Traced and tested the changes in the debugger.

-Dale

PR #34, opened 2020-11-21
Fix bogus status bar message on CMOVE indicator.

DLLarson:
Somehow a copy paste error replaced the CMOVE indicator string
with a duplicate of another string in the resource file. This
commit makes things right again.


wsu-cb (approved):
Looks good

PR #45, opened 2021-04-11
Have CGbxProjView::DoBoardClone() call CGamDoc::SetLoadingVersion()

### LINKED ISSUE #44: CGbxProjView::DoBoardClone() may fail
Important:  Start with GM not running

1)  Open GM
2)  Create a new .gbx
3)  Create a new board
4)  Clone the new board

--> The cloned board is not created.

The problem is that CGamDoc::SetLoadingVersion() isn't called until a .gbx is loaded.  If no .gx is loaded before cloning a board, the Serialize code used by the clone process doesn't know the correct file version.

wsu-cb:
Otherwise, CGbxProjView::DoBoardClone() will fail if no other
.gbx was loaded previously.

Fix #44

### DIFF
--- a/GM/GmDoc.cpp
+++ b/GM/GmDoc.cpp
@@ -614,7 +614,8 @@ void CGamDoc::Serialize(CArchive& ar)
                     AfxThrowArchiveException(CArchiveException::genericException);
                 }
             }
-            SetLoadingVersion(NumVersion(verMajor, verMinor));
+            SetLoadingVersionGuard setLoadingVersionGuard(NumVersion(verMajor, verMinor),
+                                                            NumVersion(fileGbxVerMajor, fileGbxVerMinor));
             ar >> bucket;           // Eat program version
             ar >> bucket;
 
@@ -681,8 +682,6 @@ void CGamDoc::Serialize(CArchive& ar)
 
             if (CGamDoc::GetLoadingVersion() >= NumVersion(2, 90))// Ver 2.90
                 CColorPalette::CustomColorsSerialize(ar, m_pCustomColors);
-
-            SetLoadingVersion(NumVersion(fileGbxVerMajor, fileGbxVerMinor));
         }
         CATCH(CArchiveException, e)
         {
--- a/GM/GmDoc.h
+++ b/GM/GmDoc.h
@@ -47,6 +47,10 @@
 #include    "MapStrng.h"
 #endif
 
+#ifndef     __VERSIONS_H__
+#include    "Versions.h"
+#endif
+
 //////////////////////////////////////////////////////////////////////
 
 #define     KEY_PASS_POSTFIX        "BaDkArMa"
@@ -181,6 +185,7 @@ public:
     // Current Tile Manager. Only valid when Serializing
     static CTileManager* GetCurrentTileManager() { return c_pTileMgr; }
     static void SetLoadingVersion(int ver) { c_fileVersion = ver; }
+    using SetLoadingVersionGuard = ::SetLoadingVersionGuard<CGamDoc>;
     static int GetLoadingVersion() { return c_fileVersion; }
     // -------- //
     DWORD GetGameBoxID() { return m_dwGameID; }
--- a/GM/VwPrjgb1.cpp
+++ b/GM/VwPrjgb1.cpp
@@ -146,6 +146,7 @@ void CGbxProjView::DoBoardClone()
     CBoard* pNewBoard = NULL;
     TRY
     {
+        CGamDoc::SetLoadingVersionGuard setLoadingVersionGuard(NumVersion(fileGbxVerMajor, fileGbxVerMinor));
         CMemFile file;
         CArchive arSave(&file, CArchive::store);
         arSave.m_pDocument = pDoc;
--- a/GP/GamDoc.h
+++ b/GP/GamDoc.h
@@ -65,6 +65,10 @@
 #include    "MoveMgr.h"
 #endif
 
+#ifndef     __VERSIONS_H__
+#include    "Versions.h"
+#endif
+
 ////////////////////////////////////////////////////////////////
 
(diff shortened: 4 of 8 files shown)


wsu-cb:
I have rebased this to master c0e4ce0, so it ready for merge.

DLLarson:
>I have rebased this to master c0e4ce0, so it ready for merge.

>Would this class be useful in other areas of the code?

So is this a no?
-Dale

wsu-cb:
> > I have rebased this to master [c0e4ce0](https://github.com/CyberBoardPBEM/cbwindows/commit/c0e4ce0fd952833640fa17b174ca135b41944e71), so it ready for merge.
> 
> > Would this class be useful in other areas of the code?
> 
> So is this a no?
> -Dale

I don't see the original message with the "other areas" question.  Could you please post a link to the question?

Assuming this is talking about setting and restoring the serialization version, I suspect restoring the serialization version is unnecessary.  When I posted this, I was actually wondering if you were going to recommend simplifying the code by just setting the version and not bothering to restore it.

DLLarson:
>I was actually wondering if you were going to recommend simplifying the code by just 
>setting the version and not bothering to restore it.

The version steering stuff is a bit of a hack to help steer loading of older files through the de-serialization process. I wouldn't mess with it without a thorough understating of the ramifications with backwards compatibility. I hate how brittle the whole binary serialization scheme is and how it evolves--but it works.

Perhaps one day a more flexible system based on other techniques could be introduced. However, CB should always be able to load prior file versions.

-Dale

wsu-cb:
> The version steering stuff is a bit of a hack to help steer loading of older files through the de-serialization process. I wouldn't mess with it without a thorough understating of the ramifications with backwards compatibility. I hate how brittle the whole binary serialization scheme is and how it evolves--but it works.
> 
> Perhaps one day a more flexible system based on other techniques could be introduced. However, CB should always be able to load prior file versions.
> 
> -Dale

For what it's worth, the binary serialization system doesn't bother me.  Indeed, I far prefer it to pseudo-text schemes like .xml, .json, etc.  (As a minor issue, object-oriented purity would probably dictate that the version be an instance member of the CGamDoc being serialized instead of a class-global.  On the other hand, that change would complicate this technique of cloning a board by serialization without a CGamDoc object.)

But to go back to https://github.com/CyberBoardPBEM/cbwindows/pull/45#issuecomment-818907383:  I don't know what your question about "other areas" was referring to, so I don't know whether the answer to the question is yes or no.

DLLarson:
> I don't know what your question about "other areas" was referring to, so I don't know 
>whether the answer to the question is yes or no.

I was thinking that there were several other areas in the code that bracket some data loads with a temporary file version change call to `SetLoadingVersion` that would then restore the original version after a series of calls or scope end. 

So it seemed like such a class would be useful in other areas. Clearly there is nothing wrong with the code and we can let it rip as is.

Basically a suggestion.

-Dale

wsu-cb:
> I was thinking that there were several other areas in the code that bracket some data loads with a temporary file version change call to `SetLoadingVersion` that would then restore the original version after a series of calls or scope end.
> 
> So it seemed like such a class would be useful in other areas. Clearly there is nothing wrong with the code and we can let it rip as is.
> 
> Basically a suggestion.
> 
> -Dale

OK, now I understand.  I searched through the code just now, and do see save/restore logic elsewhere.  I can work on a new version of this PR that applies this concept where the current code saves and restores the version.

However, I see places that appear to set a version at the beginning of a function, and a different version at the end of the function, but not necessarily the original version.  For example, see CGamDoc::SerializeMoveSet().  Is that intentional, or should that actually be changed to a version restore?  You stated above that the serialization logic is somewhat brittle, so I won't change the current semantics of the code without a clear understanding of the intent.

wsu-cb:
More serialization questions:

I'm looking at `CGameBox::Load()`, and the clean up at the end of the function (`c_pTileMgr`, `CGamDoc::SetLoadingVersion()`) doesn't get executed in the exception handler cases.  Would I be correct in thinking those actions should also be done when exceptions occur?

What is the distinction between `CGamDoc::SetLoadingVersion()` and `CGameBox::SetLoadingVersion()`?  I assume the first is for .gam and the second for .gbx files, but can a .gam ever be (successfully) loaded if its version is different from the .gbx?  If not, why are there two distinct version values for these classes (rather than a single, shared, version)?  If so, why does `CGameBox::Load()` call `CGamDoc::SetLoadingVersion()` with the .gbx version, thus apparently ignoring the .gam version?

There are places where it looks like the `SetLoadingVersion()` is used at the end of a function without an obvious initial setting, so it doesn't look like a restore operation.  For example, `CMoveList::PopAndRestoreState()` and `CGamDoc::Serialize()`.  What is the intent in these cases?

DLLarson:
Hi,

>You stated above that the serialization logic is somewhat brittle, 
>so I won't change the current semantics of the code without a clear understanding of the intent.

Yep... I wouldn't change the semantics at all. Even if not clear.

> For example, CMoveList::PopAndRestoreState() and CGamDoc::Serialize(). What is the intent in these cases?

Most times it's to temporarily match loader code with the vintage of loading file data. Looking back I really don't recall what's going on in this particular case. Best to leave it be. :)

-Dale

DLLarson:
>I'm looking at CGameBox::Load(), ... doesn't get executed in the exception handler cases...
>Would I be correct in thinking those actions should also be done when exceptions occur?

It would probably be proper to balance the sets with restores. In this case I believe the entire load is abandoned with an error message so it shouldn't matter. I think it could be loading a GameBox file that may not match the Game/Scenario versions being loaded.

-Dale

DLLarson (approved):
Tested against random mix of older version gameboxes and games.

I'm really liking the piecewise modernizing of the code! It should really help reliability.

-Dale

PR #47, opened 2021-04-12
Have CDib::DIBToBitmap return std::unique_ptr<>

### LINKED ISSUE #46: CBitmapImage::CopyAttributes() memory leak
There is a memory leak in CBitmapImage::CopyAttributes().  This can be seen by loading the attached .gsn into GP, closing GP, and then looking at the MSVC memory leak report.  (The memory is not allocated by CBitmapImage::CopyAttributes(), but that's the function that should release the memory.)

[geo001.zip](https://github.com/CyberBoardPBEM/cbwindows/files/6292935/geo001.zip)

wsu-cb:
instead of bare pointer to fix CBitmapImage::CopyAttributes()
memory leak.  Note that this commit is changing code everywhere
except where the original code is wrong, but is in line with the
long term goal of modernizing the code base.

Fix #46

### DIFF
--- a/GShr/CDib.cpp
+++ b/GShr/CDib.cpp
@@ -155,7 +155,7 @@ BOOL CDib::BitmapToDIB(CBitmap* pBM, CPalette* pPal, int nBPP/* = 16*/)
     return m_hDib != NULL;
 }
 
-CBitmap* CDib::DIBToBitmap(CPalette *pPal, BOOL bDibSect /* = TRUE */)
+std::unique_ptr<CBitmap> CDib::DIBToBitmap(CPalette *pPal, BOOL bDibSect /* = TRUE */)
 {
     if (bDibSect)
     {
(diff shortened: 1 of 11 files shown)


wsu-cb:
Please note that I have now sent two PRs against the same master commit.  We will need to rebase one of them after the other is applied.

GShr/CDib.cpp
```diff
@@ -155,7 +155,7 @@ BOOL CDib::BitmapToDIB(CBitmap* pBM, CPalette* pPal, int nBPP/* = 16*/)
     return m_hDib != NULL;
 }
 
-CBitmap* CDib::DIBToBitmap(CPalette *pPal, BOOL bDibSect /* = TRUE */)
+std::unique_ptr<CBitmap> CDib::DIBToBitmap(CPalette *pPal, BOOL bDibSect /* = TRUE */)
```
DLLarson, line 158: I admit I'm still getting used to the efficient returning of C++ objects in the newer standards!

My knee-jerk reaction is to be suspicious of the overhead.

-Dale

DLLarson (approved):
Looks good to me.
-Dale

Issue #48, opened 2021-04-25
problem when changing CTrayPalette tray name with keyboard


wsu-cb:
1)  Open a .gam with multiple piece trays
2)  Open a tray palette
3)  Using the mouse, drop down the tray palette's list of names
4)  Using the keyboard arrow keys, change the selected name, and hit Enter

--> The drop down closes, the tray palette name shows the selected name, but the displayed pieces are from the old piece tray

I see that `CTrayPalette::OnTrayNameCbnSelchange()` explicitly does not update the tray list when the drop down is open, but that causes this problem.  (If I remove the blocking `if` test, the correct pieces are displayed.)

With the `if` test disabled, I see that updating the tray piece display causes changing the selected name with the keyboard arrows to be very slightly sluggish, so not updating the piece display may be intentional.  If that is the case, then I think we need to add handling for `ON_CBN_SELENDOK` to update the piece display. However, my preference is to make the piece display correspond to the focused name despite the slight delay.  (This may be a case of modern computers' better performance making some UI sacrifices unnecessary.)

Is making the piece display static while using the keyboard arrows intentional?

DLLarson:
>Is making the piece display static while using the keyboard arrows intentional?

No. It's definitely a bug.

-Dale

wsu-cb:
> > Is making the piece display static while using the keyboard arrows intentional?
> 
> No. It's definitely a bug. Closing the dropdown should've detected the selection change.
> 
> -Dale

I'm afraid this answer didn't clarify the intended behavior for me.  The question I was trying to ask was whether the piece display should be changed in the `ON_CBN_SELCHANGE` handler or in the `ON_CBN_SELENDOK` handler.  The 'No' part of your answer sounds like updating the display `ON_CBN_SELCHANGE`, but the 'selection change' sounds like updating the display `ON_CBN_SELENDOK`.

DLLarson:
Ok... I see more clearly what you are asking. I just tried the program after removing the `if` and it now updates changing the selection with the arrow keys. I don't really have any lags but my machine is pretty new.

>However, my preference is to make the piece display correspond to the focused name despite the slight delay. 

I think this is the correct approach because it tracks with what the dropdown does if you use the arrow keys when the dropdown isn't visible.

-Dale

Issue #43, opened 2021-04-04
CPlayBoardView::DoDragSelectList() HINT_UPDATESELECTLIST


wsu-cb:
In `CPlayBoardView::DoDragSelectList()`, a `HINT_UPDATESELECTLIST` message is sent with no hint object.  However, a hint object is constructed immediately before the message is sent, but then never actually used.  Should that hint object be sent as the argument for `HINT_UPDATESELECTLIST`, or should the hint object be removed?

For what it's worth, I suspect the hint in `CPlayBoardView::DoDragSelectList()` should be removed because I see another `HINT_UPDATESELECTLIST` sent with no argument, and also the only code that processes `HINT_UPDATESELECTLIST` does not look for a hint argument.

Issue #41, opened 2021-02-28
possible extraneous member variable CMarkListBox::m_bDisplayIDs


wsu-cb:
If I change `CMarkListBox::m_bDisplayIDs` to type `const int`, the compiler reports no errors for attempting to assign to it.  Therefore, I think it is currently unused.

Since `CMarkListBox` is derived from `CTileBaseListBox`, and there is also declared `CTileBaseListBox::m_bDisplayIDs`, which is in use, I suspect that `CMarkListBox::m_bDisplayIDs` should be removed.

Does this appear correct?

DLLarson:
>Therefore, I think it is currently unused.

Actually it is used. This flag is set by the `CTileListBox` constructor by reading the registry:

```
    m_bDisplayIDs = AfxGetApp()->GetProfileInt("Settings", "DisplayIDs", 0);
```

It's used for diagnosis of various ID's (Tiles, etc...) by showing the ID's in the UI to compare against other ID dump tools hidden in the code.

I think the reason it didn't trip a compiler error is because it was set in the constructor of the containing class.

-Dale

wsu-cb:
Doesn't `CTileListBox`'s constructor set `CTileListBox::m_bDisplayIDs`?  As a derived class, I don't think `CTileListBox`'s constructor can reach `CMarkListBox::m_bDisplayIDs`.

If `CMarkListBox::m_bDisplayIDs` is used for a different purpose than `CTileListBox::m_bDisplayIDs`, why does it have the same name?

On the other hand, if it is used for the same purpose, why can't it just use `CTileListBox::m_bDisplayIDs` (since the base class version is protected, and thus accessible to the derived class)?

DLLarson:
>On the other hand, if it is used for the same purpose, why can't it just 
>use CTileListBox::m_bDisplayIDs (since the base class version is protected, and 
>thus accessible to the derived class)?

My bad. 

I missed the specificity of the `CMarkListBox` class. That `CMarkListBox::m_bDisplayIDs` member appears to be a copy/paste code cloning remnant. It's dead code as you indicate.

Also, my explanation in an earlier post...

>It's used for diagnosis of various ID's (Tiles, etc...) 

...needs to be refined. It shows the Tile ID's only for various elements in the GUI. I used it to run down some tile reference issue a long time ago along with a hidden hotkey to dump the TileID's related internal tables. I generally like to leave simple tools like this in "just-in-case".

Sorry for the diversion.

-Dale

PR #51, opened 2021-05-14
Tray kbd update

### LINKED ISSUE #37: CTileSheet::ClearSheet() deletes non-heap m_pDC
CTileSheet::ClearSheet() calls `delete m_pDC`.  However, my reading of the code says `m_pDC` never contains a heap-allocated object.  Fortunately, my reading of the `CTileSheet` code is that `CTileSheet::ClearSheet()` is never called with `m_pDC != NULL`.

Nevertheless, I think the code should be rewritten to reflect the current behavior.

(Actually, the way I read the code, I think m_pDC should be removed from the class entirely.  It looks to me like it's only used as a transient value, and should only be stored on the stack.)

wsu-cb:
Combine some small updates into a single PR

Fix #37 #41 #43 #48

### DIFF
--- a/GP/PalTray.cpp
+++ b/GP/PalTray.cpp
@@ -758,8 +758,7 @@ void CTrayPalette::OnWindowPosChanging(WINDOWPOS FAR* lpwndpos)
 
 void CTrayPalette::OnTrayNameCbnSelchange()
 {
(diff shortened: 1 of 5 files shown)


DLLarson (approved):
Look good to me.

Thx!

-Dale

PR #52, opened 2021-06-27
Propconst+notnull+ptrlist

wsu-cb:
Here's some more type safety work.

1)  http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4388.html explains an often-undesirable characteristic of the C/C++ const type qualifier, and proposes a std::experimental::propagate_const<> template to provide a way to get around this.  However, this is not widely available, so I add a CB::propagate_const<> template to approximate the proposed standard, and then make use of it.

2)  In many cases, pointers are expected not to be null.  Add a CB::not_null<> template that can perform this checking automatically, and assert/throw if this assumption is violated.  Then make use of CB::not_null.

3)  Replace uses of MFC CPtrList (which contains void*) with more narrowly typed uses of std:;list<>.

### DIFF
--- a/GShr/CyberBoard.h
+++ b/GShr/CyberBoard.h
@@ -354,6 +354,38 @@ namespace CB
     }
 }
 
+template<typename T>
+using OwnerPtr = CB::not_null<CB::propagate_const<std::unique_ptr<T>>>;
+
+// like c++14 std::make_unique<>
+template<typename T, typename... Args>
+OwnerPtr<T> MakeOwner(Args&&... args)
+{
+    return OwnerPtr<T>(new T(std::forward<Args>(args)...));
+}
+
+template<typename T>
+class OwnerOrNullPtr : public CB::propagate_const<std::unique_ptr<T>>
+{
+    using BASE = CB::propagate_const<std::unique_ptr<T>>;
+
+public:
+    using BASE::BASE;
+    using BASE::operator=;
+
+    OwnerOrNullPtr() = default;
+
+    template<typename U>
+    OwnerOrNullPtr(OwnerPtr<U>&& pu) : BASE(CB::get_underlying(std::move(pu))) {}
+
+    template<typename U>
+    OwnerOrNullPtr& operator=(OwnerPtr<U>&& pu)
+    {
+        *this = CB::get_underlying(std::move(pu));
+        return *this;
+    }
+};
+
 /////////////////////////////////////////////////////////////////////////////
 
 /* Invalid_v<T> is used as the "no-value" value for type T.
--- a/GShr/CyberBoard.h
+++ b/GShr/CyberBoard.h
@@ -59,15 +59,6 @@ namespace CB
         }
     }
 
-    template<typename RIGHT>
-    void Copy(CPtrList& left, const RIGHT& right)
-    {
-        left.RemoveAll();
-        for (RIGHT::const_iterator it = right.begin() ; it != right.end() ; ++it) {
-            left.AddTail(&**it);
-        }
-    }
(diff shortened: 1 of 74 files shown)


DLLarson:
You've been busy! :)

This one--like the earlier large PR--will take a few days to get through.

-Dale

DLLarson (approved):

DLLarson:
At first I thought I'd try a different tact of inspecting the individual commits but it seems like when I OK the commit it applies to the full PR. This seems confusing. I'll keep doing this but not mark the commits as inspected until I the very end when I'll do it in aggregate.

-Dale

GShr/CyberBoard.h
```diff
@@ -79,6 +79,136 @@ namespace CB
     }
 }
```
DLLarson, line 72: I'm assuming the plan is to remove this code when MSVC adopts the proposal?

GShr/CyberBoard.h
```diff
@@ -354,6 +354,38 @@ namespace CB
...
 
+template<typename T>
+using OwnerPtr = CB::not_null<CB::propagate_const<std::unique_ptr<T>>>;
+
```
DLLarson, line 333: Inconveniently verbose indeed!

GShr/CyberBoard.h
```diff
@@ -99,8 +99,21 @@ constexpr T Invalid_v = Invalid<T>::value;
 
 /////////////////////////////////////////////////////////////////////////////
 
+/* TODO:  The checks could be removed to improve release
```
DLLarson, line 430: Why trust when you don't have to? I say keep the test.

GShr/DrawObj.h
```diff
@@ -649,6 +649,12 @@ class CMarkObj : public CDrawObj
 ///////////////////////////////////////////////////////////////////////
 // The DrawList manager class
 
+/* WARNING:  Ownership of objects in CDrawList is hard to
```
DLLarson, line 650: Interesting!...

I suppose this should be attended to at some point. Thanks for adding the note.

GShr/CyberBoard.h
wsu-cb: Yes.

wsu-cb:
> At first I thought I'd try a different tact of inspecting the individual commits but it seems like when I OK the commit it applies to the full PR. This seems confusing. I'll keep doing this but not mark the commits as inspected until I the very end when I'll do it in aggregate.
> 
> -Dale

Should I wait to push updates to address your comments (e.g., const CPalette*, thorough self-checking) until after you have finished reviewing the initial PR?

DLLarson:
>Should I wait to push updates to address your comments (e.g., const CPalette*, thorough self-checking) until after you have finished reviewing the initial PR?

Yes. Best to wait. I don't think there'll be much to do.

-Dale

DLLarson (changes requested):
The only possible changes would be the addition of `const` to the palette pointers.

Code seems to operate fine when cursory tested against a variety of gamebox/games.

-Dale

DLLarson (approved):

DLLarson:
I'm conflicted on how to do the merge: collapse to a single commit or keep all the commits. 

Since it's your PR I'll defer to you for the choice.

-Dale

wsu-cb:
> I'm conflicted on how to do the merge: collapse to a single commit or keep all the commits.
> 
> Since it's your PR I'll defer to you for the choice.
> 
> -Dale

My preference is to keep all the commits.  I think it's easier to analyze what's going on when commits are smaller.

PR #55, opened 2021-07-04
select pieces when dropping the into Piece Tray

### LINKED ISSUE #54: pieces dropped in Piece Tray not selected
I expect pieces that are dropped into a Piece Tray to be selected, but that does not happen, both when the user is directly dropping pieces, and when playback shows pieces being dropped.

wsu-cb:
This PR selects pieces when they are dropped into a Piece Tray, both directly and during playback.

However, note that each time playback prepares to show the next move, all selections are cleared.  This means "Automatic Playback" effectively prevents Piece Tray selection from being visible.  That appears to be a pretty deeply embedded assumption, and may be considerable work to change.

Fix #54


DLLarson (changes requested):
I see one issue. If I drag a selected group of tray pieces into another tray the pieces in the new tray aren't selected.
-Dale

DLLarson (approved):
Tested tray to tray, board to tray, tray to board, and intertray drags. All good!

-Dale

PR #56, opened 2021-07-11
Type safety update for GameElement

### LINKED ISSUE #53: GameElement special values 0 vs -1
I see that `CGamDoc::GetGameElementCodeForObject()` returns `(GameElement)0` for inappropriate inputs, but 'CGamDoc::GetVerifiedGameElementCodeForObject()' returns `(GameElement)-1`.  Why the difference?

For what it's worth, it looks to me like `(GameElement)0` would be the result of `MakePieceElement(PieceID(0))`, which suggests to me that `(GameElement)0` isn't a good "error" value because it's actually valid in some cases.  On the other hand, MarkID(0xFFFF) should never be valid, so `(GameElement)-1` looks like an appropriate "error" value.

In short, could we replace uses of `(GameElement)0` with `(GameElement)1`?

wsu-cb:
Make the precise layout of GameElement more explicit in preparation for increasing ID width.

Fix #53

### DIFF
--- a/GShr/DrawObj.h
+++ b/GShr/DrawObj.h
@@ -189,7 +189,13 @@ protected:
 class alignas(uint32_t) ObjectID
 {
 public:
-    ObjectID();
+    // KLUDGE:  release link fails if impl is in .cpp
+    constexpr ObjectID() :
+        id(0),
+        serial(0),
+        subtype(0)
+    {
+    }
     ObjectID(uint16_t i, uint16_t s, CDrawObj::CDrawObjType t);
     explicit ObjectID(PieceID pid);
     explicit ObjectID(uint32_t dw);
--- a/GShr/DrawObj.h
+++ b/GShr/DrawObj.h
@@ -193,6 +193,9 @@ public:
     ObjectID(uint16_t i, uint16_t s, CDrawObj::CDrawObjType t);
     explicit ObjectID(PieceID pid);
     explicit ObjectID(uint32_t dw);
+    ObjectID(const ObjectID&) = default;
+    ObjectID& operator=(const ObjectID&) = default;
+    ~ObjectID() = default;
 
     bool operator==(const ObjectID& rhs) const
     {
--- a/GShr/DrawObj.h
+++ b/GShr/DrawObj.h
@@ -192,7 +192,7 @@ public:
     ObjectID();
     ObjectID(uint16_t i, uint16_t s, CDrawObj::CDrawObjType t);
     explicit ObjectID(PieceID pid);
-    explicit ObjectID(DWORD dw);
+    explicit ObjectID(uint32_t dw);
 
     bool operator==(const ObjectID& rhs) const
     {
@@ -201,6 +201,12 @@ public:
     bool operator!=(const ObjectID& rhs) const
     {
         return !operator==(rhs);
+    }
+
+    explicit operator uint32_t() const
+    {
+        static_assert(sizeof(uint32_t) == sizeof(*this), "sizeof error");
+        return reinterpret_cast<const uint32_t&>(*this);
     }
 
 private:
--- a/GShr/MapStrng.cpp
+++ b/GShr/MapStrng.cpp
@@ -27,6 +27,19 @@
 
 ///////////////////////////////////////////////////////////////////////
 
+namespace {
+    class GameElementCheck
+    {
+    public:
+        GameElementCheck()
+        {
+            GameElement test(MarkID(0x1234));
+            static_assert(sizeof(test) == sizeof(uint32_t), "size mismatch");
+            ASSERT(reinterpret_cast<uint32_t&>(test) == 0xF0001234 || !"non-Microsoft field layout");
+        }
+    } gameElementCheck;
+}
+
 void CGameElementStringMap::Clone(CGameElementStringMap* pMapToCopy)
 {
     RemoveAll();
@@ -76,7 +89,7 @@ void CGameElementStringMap::Serialize(CArchive& ar)
             GameElement elem;
             CString str;
             GetNextAssoc(pos, elem, str);
-            ar << (DWORD)elem;
+            ar << elem;
             ar << str;
         }
     }
@@ -87,11 +100,11 @@ void CGameElementStringMap::Serialize(CArchive& ar)
         ar >> dwCount;
         while (dwCount--)
         {
-            DWORD dwElem;
+            GameElement dwElem;
             ar >> dwElem;
             CString str;
             ar >> str;
-            SetAt((GameElement)dwElem, str);
+            SetAt(dwElem, str);
         }
     }
 }
--- a/GShr/MapStrng.cpp
+++ b/GShr/MapStrng.cpp
@@ -27,6 +27,8 @@
 
 ///////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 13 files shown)


GShr/DrawObj.h
```diff
@@ -189,10 +189,19 @@ class CDrawObj
 class alignas(uint32_t) ObjectID
 {
 public:
-    ObjectID();
+    // KLUDGE:  release link fails if impl is in .cpp
```
DLLarson, line 192: Any idea why the link fails in this case? Does it seem like a compiler bug?

GShr/MapStrng.cpp
```diff
@@ -27,6 +27,19 @@
 
 ///////////////////////////////////////////////////////////////////////
 
+namespace {
+    class GameElementCheck
```
DLLarson, line 31: What's this about? Is this some left over test code?

GShr/MapStrng.h
```diff
@@ -38,42 +38,326 @@
 
 //////////////////////////////////////////////////////////////////////
 
-typedef DWORD GameElement;
+typedef DWORD GameElementOld;
```
DLLarson, line 41: I'm confused about the role of `GameElementOld`. Sometimes it's used on debug builds. Sometimes not. Perhaps it's the name of the typedef.

DLLarson (changes requested):
Some questions are "queued".

-Dale

GShr/DrawObj.h
wsu-cb: At first glance, it kind of feels like a compiler bug.  However, I can actually imagine some reasons why forced inlining of a constexpr function makes sense.  I can do some experiments and see if a simpler example results in similar behavior, but it probably won't be today.

GShr/MapStrng.cpp
wsu-cb: Test code, yes, but not leftover.  I wanted to leave this in long-term since other compilers and systems may use different bit layouts.

GShr/MapStrng.h
wsu-cb: The `GameElementOld` code is intended to be used consistently in debug builds to verify that the new code produces`GameElement` results that are binary-equal with the original code.  Where do you see it not being used in debug builds?

Since `ASSERT` doesn't do anything in release builds, I didn't see any reason to have any of the `GameElementOld` code run in release builds.

Unlike `GameElementCheck`, which is intended to check that files can be exchanged with new system ports, I expect to remove this code after a test period.

GShr/MapStrng.cpp
DLLarson: Ok. Perhaps a comment about its purpose and longevity would've helped.

GShr/MapStrng.h
```diff
@@ -38,42 +38,326 @@
...
+        GameElement retval = GameElement(mid);
+#if !defined(NDEBUG)
+        GameElementOld old = MakeMarkerElementOld(mid);
+        ASSERT(reinterpret_cast<GameElementOld&>(retval) == old);
+#endif
+        return retval;
+    }
 
 #if defined(GPLAY)
+inline GameElementOld MakeObjectIDElementOld(ObjectID dwObjectID)
```
DLLarson, line 291: This `GameElementOld` appears to be in the release build.

GShr/MapStrng.h
```diff
@@ -38,42 +38,326 @@
...
+#if !defined(NDEBUG)
+inline GameElementOld MakeMarkerElementOld(MarkID mid)
+    { return (GameElementOld)((DWORD)static_cast<WORD>(mid) | GAMEELEM_MARKERID_FLAG); }
+#endif
 inline GameElement MakeMarkerElement(MarkID mid)
-    { return (GameElement)((DWORD)static_cast<WORD>(mid) | GAMEELEM_MARKERID_FLAG); }
+    {
+        GameElement retval = GameElement(mid);
+#if !defined(NDEBUG)
+        GameElementOld old = MakeMarkerElementOld(mid);
```
DLLarson, line 284: This `GameElementOld` appears to be in the release build.

GShr/MapStrng.h
DLLarson: This definition `GameElementOld` appears to be in the release build.

DLLarson, line 41: I think I'm really just having trouble wrapping my brain around the name. Perhaps something to the effect `GameElementLegacyCheck` or some such to capture its role rather than `Old`?

GShr/DrawObj.h
DLLarson: Ok... no worries here.

GShr/MapStrng.cpp
wsu-cb: I'm afraid I thought the "non-Microsoft field layout" text explained this.  However, I will add a more-explicit comment.

GShr/MapStrng.h
wsu-cb: Oops.  You're right; I missed that one.

GShr/MapStrng.h
wsu-cb: And I missed that one.

GShr/MapStrng.h
wsu-cb: But I think this one is (as intended) only in debug.

GShr/MapStrng.h
wsu-cb: OK

GShr/DrawObj.h
wsu-cb: (Draft) C++17 standard section 10.1.5 says constexpr functions are implicitly inline.  Therefore, it's more surprising when compilation succeeds rather than fails with the implementation in the .cpp.  Indeed, on a very simple project test, the linker failed in both debug and release builds when the implementation was in the .cpp.

So the real question is why does the Cyberboard debug link succeed?

But at any rate, I think the function does need to be inline.

DLLarson:
I ran into an ASSERT while testing the changes:

image

To replicate use the attached archive.

[War1812-ASSERT-duringplayback-TestFiles.zip](https://github.com/CyberBoardPBEM/cbwindows/files/6805452/War1812-ASSERT-duringplayback-TestFiles.zip)

Load the game into the debug session of cbplay and do this:

image

-Dale

wsu-cb:
> I ran into an ASSERT while testing the changes:
> 

I see the problem, and how to fix it.  However, it's not related to this PR; this mistake has been (unnoticed) in master for a while.  Should I add the fix for this problem to this PR, or should I create a separate PR to fix this problem?

DLLarson:
>Should I add the fix for this problem to this PR, or should I create a separate PR to fix this problem?

Let's put it into a different PR then rather than cloud up this commit. 

l'll merge this one in. Would you prefer I retain all the commits or squash them with the squashed commit containing a roll-up of all the commits in its comment? i.e., Is there any value in this case with keeping the "sausage making"?

wsu-cb:
> l'll merge this one in. Would you prefer I retain all the commits or squash them with the squashed commit containing a roll-up of all the commits in its comment? i.e., Is there any value in this case with keeping the "sausage making"?

This is small enough that squashing seems reasonable.

DLLarson (approved):
Good to go!
-Dale

Issue #57, opened 2021-07-11
ObjectID subtype interpretation


wsu-cb:
After previous discussions, I described the top 4 bits of `ObjectID` as a subtype field where value 0 represents the "invalid" subtype.  However, after looking at `ObjectID` again, I am noticing that `ObjectID`'s ID field can contain a `PieceID` (instead of a `CMarkObj`'s random number), and that the subtype is 0 in this case.  So would it be better to describe subtype 0 as representing a `PieceID` instead of invalid?

DLLarson:
I'm still thinkin' on this...

wsu-cb:
> I'm still thinkin' on this...

Thank you for letting me know it hasn't been overlooked.

wsu-cb:
To add to the headache here, I now realize that tag v3.5-prerelease-2 has `CDrawObj::GetObjectID()` return `0` as if that is the "invalid" ObjectID, but it looks to me like `CPieceObj::GetObjectID()` will also return `0` when `CPieceObj` wraps `PieceID(0)`.  That suggests to me that it would be better to use `DWORD(PieceID(0xffff))` as the invalid ObjectID since that can't be mistaken for a valid Piece's ObjectID.

(Some of the type safety work I have already done is intended to make the ObjectID construction more explicit, but the actual final bit patterns should be unchanged by that work.)

DLLarson:
Hi Bill,

>That suggests to me that it would be better to use DWORD(PieceID(0xffff)) as 
>the invalid ObjectID since that can't be mistaken for a valid Piece's ObjectID.... 
>but the actual final bit patterns should be unchanged by that work.

I agree. I've been trying to historically wrap my brain around the whole scheme's evolution and I just can't seem to remember many the particulars that drove a decision. 

Most of the "packed" data items are really the result of Cyberboard originally running on Win16 systems which had extreme limits on memory (real mode memory footprints), speed (plus no floating point), graphics (256 color then 16bit color) and even the number of font's that could be "realized" into memory at a given moment (hence the use of "atoms" to prevent redundant font realizations.) Much of these are no longer true limitations on modern platforms. 

So this evolving of the codebase is really important to surface, segregate and compartmentalize those areas that had limitations so that CB can better take advantage of modern capabilities without breaking past work.

Thanks for your keen insight and applying your skills to ferret out the edge cases!

-Dale

wsu-cb:
[...]

PR #61, opened 2021-08-11
file version command line option /id32

### LINKED ISSUE #59: file upgrade warning logic
If I am reading the code correctly, both GM and GP only warn about upgrading a file if the file is older than version 2.90.  However, the file will still be upgraded for versions higher than that, but less than the current version.

Am I misunderstanding the code, or is there a reason the warning isn't displayed for version >= 2.9, or is this a bug (i.e., there should be a warning for any version less than the current one)?

wsu-cb:
This implements the file version command line option /id32 discussed in #57.

Also Fix #59

### DIFF
--- a/GM/Gm.cpp
+++ b/GM/Gm.cpp
@@ -127,7 +127,17 @@ CGmApp::CGmApp()
 /////////////////////////////////////////////////////////////////////////////
 // The one and only CGmApp object
 
-CGmApp theApp;
+CGmApp& CGmApp::Get()
+{
+    /* make this a function variable to ensure
+        it's constructed before use */
+    static CGmApp theApp;
+    return theApp;
+}
+CWinApp& CbGetApp()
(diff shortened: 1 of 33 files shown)


DLLarson:
Hi Bill,

I've seen this and I plan to look this over tomorrow (8/13).

-Dale

DLLarson (changes requested):
Hi Bill,

Looks good to me.

>Note:  this commit throws an exception when saving with
file v 4.00, so will not produce any invalid files.  Therefore,
it might make sense to squash this commit with the previous one.

I agree. Leave the first two commits alone and squash the last two. The last two are a unit of change. I'll leave that rebase to you before merging the final set into master.

-Dale

DLLarson:
Looking good!
-Dale

DLLarson (approved):

PR #62, opened 2021-08-21
use GameElement for CGrafixListBox ItemCode

wsu-cb:
The `CGrafixListBox* *ItemCode*` functions store their item descriptions in an `int`.  However, I think `GameElement` already provides storage for everything `ItemCode` needs to store, and adding very simple accessors to `GameElement` allows `ItemCode` users to extract the data they need.

Since `GameElement` will need to be adjusted to handle 32bit IDs anyway, refactoring the code this way will save having to also add 32bit ID handling to the `ItemCode` functions.

Am I overlooking some reason the `ItemCode` functions shouldn't use `GameElement`?

### DIFF
--- a/GM/VwBitedt.h
+++ b/GM/VwBitedt.h
@@ -59,7 +59,7 @@ public:
         { return CSize(m_size.cx * m_nZoom, m_size.cy * m_nZoom); }
     CRect GetZoomedSelectBorderRect();
     CRect GetZoomedSelectRect();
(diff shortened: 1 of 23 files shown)


DLLarson:
Hi Bill,
I've seen this PR. I'll be looking at it in detail tomorrow (8/24.)
-Dale

DLLarson (approved):
Looks good to me!

I don't see any issues with the shift from `int` to the explicit type.

-Dale

PR #68, opened 2021-09-19
fix some missing initialization and an incorrect ASSERT

wsu-cb:
I found and fixed some bugs when I tried to create and save a new .gbx today.

These two commits aren't really related to each other, except that I found both problems today.


DLLarson (approved):
Looks good to me.

Merge without squash?

wsu-cb:
> Merge without squash?

Right

PR #69, opened 2021-09-19
Fix #67

### LINKED ISSUE #67: possible bug in Rotate.cpp:CalcRotatedRect()
The code in Rotate.cpp:CalcRotatedRect() looks a bit odd:
```
static CSize CalcRotatedRect(CSize size, int angle, POINT* pSPnts, POINT* pDPnts)
{
...
    int xHalf = size.cx / 2;
    int yHalf = size.cy / 2;
    OffsetRectVector(pDPnts, -xHalf, -xHalf);
    RotateRectVector(pDPnts, angle);
...
}
```

This code initializes but does not use `yHalf`, and the `OffsetRectVector(pDPnts, -xHalf, -xHalf)` looks a little odd.  Should that be `OffsetRectVector(pDPnts, -xHalf, -yHalf)`?

To be clear, I have no evidence that there is an externally observable bug, but the code looks strange to me.

wsu-cb:
Fix #67

### DIFF
--- a/GP/VwPbrd.cpp
+++ b/GP/VwPbrd.cpp
@@ -1827,7 +1827,7 @@ void CPlayBoardView::DoRotateRelative(BOOL bWheelRotation)
     pos = m_tblCurPieces.begin();
     for (int i = 0 ; pos != m_tblCurPieces.end() ; i++)
     {
-        CDrawObj& pDObj = **pos;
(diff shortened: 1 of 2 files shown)


DLLarson (approved):
Looks good. I'll absorb these without squash.

-Dale

PR #78, opened 2022-03-06
Geomorphic board enhancements

wsu-cb:
I wanted to make some enhancements to the geomorphic board code, but that led me to doing more modernization work on the tile and board code.  The end result is another large PR, but I have tried to make the individual commits manageable.


GM/VwTilesl.cpp
```diff
@@ -284,16 +279,16 @@ void CTileSelView::UpdateDocumentTiles()
 {
     CGamDoc *pDoc = GetDocument();
     // Make sure our bitmaps are up to date.
-    CloneBitmap(GetActiveBitmap(), m_pEditView->GetCurrentViewBitmap());
+    GetActiveBitmap() = CloneBitmap(*m_pEditView->GetCurrentViewBitmap());
```
DLLarson, line 282: I don't understand this line of code at all after looking at the implementation. What's going on here?

-Dale

GM/VwTilesl.cpp
```diff
@@ -307,7 +302,7 @@ void CTileSelView::UpdateDocumentTiles()
 void CTileSelView::DoTileResizeDialog()
 {
     CResizeTileDialog dlg;
-    CloneBitmap(GetActiveBitmap(), m_pEditView->GetCurrentViewBitmap());
+    GetActiveBitmap() = CloneBitmap(*m_pEditView->GetCurrentViewBitmap());
```
DLLarson, line 305: Same as before. I don't understand what's going on here:

`GetActiveBitmap() = ...`

-Dale

GM/VwTilesl.cpp
```diff
@@ -356,29 +347,27 @@ void CTileSelView::DoTileResizeDialog()
 
 void CTileSelView::DoTileRotation(int nAngle)
 {
-    CDib dibFull;
-    CDib dibHalf;
-    CloneBitmap(GetActiveBitmap(), m_pEditView->GetCurrentViewBitmap());
-    if (!dibFull.BitmapToDIB(&m_bmFull, GetAppPalette()))
+    CDib dibFull(*m_bmFull, GetAppPalette());
+    CDib dibHalf(*m_bmHalf, GetAppPalette());
+    GetActiveBitmap() = CloneBitmap(*m_pEditView->GetCurrentViewBitmap());
```
DLLarson, line 352: Dittos previous two comments in this file.

DLLarson:
Hi Bill,

I'm about 2/3's through the code review. It's a lot more code than I expected! Hopefully this will be the last of the mega-commits. I see that most of the changes are just the ripple effects of changes to argument referencing and such. It's easy to get lost in all that and miss the essential changes that triggered it all.

I'm needing a break to keep my eyes from crossing. I'll finish tomorrow (3/13/22) to dig through some of the bigger changed files like the Geo stuff.

-Dale

GM/VwTilesl.cpp
wsu-cb: 1)  Commit 8a48c is about const- and null- correctness of GdiTools.h.  The original GditTools.h `CloneBitmap()` was declared `void CloneBitmap(CBitmap *pbmDst, CBitmap *pbmSrc)`.
- Applying const-correctness results in `void CloneBitmap(CBitmap *pbmDst, const CBitmap *pbmSrc)`.
- Applying null-correctness results in `void CloneBitmap(CBitmap& pbmDst, const CBitmap& pbmSrc)`.
- I now argue that the logical intent of this function is to take one input (pbmSrc) and create one output (pbmDst), so I turn that into `CBitmap CloneBitmap(const CBitmap& pbmSrc)`.
- Unfortunately, MFC does not allow copying or moving from one CBitmap to another, so the return value has to be changed to something that can be moved or copied, resulting in `CBitmap* CloneBitmap(const CBitmap& pbmSrc)`.
- Then, to get back the automatic cleanup that using a class rather than a pointer provides, that gets turned into `OwnerPtr<CBitmap> CloneBitmap(const CBitmap& pbmSrc)`.
- So the modified `CloneBitmap()` behavior is intended to be almost the same as the previous behavior.  The only difference is that the argument and return types are intended to be more precise.

2)  The `v3.5-prerelease-2` code's destination was `GetActiveBitmap()`, which was declared in VwTilesl.h as `CBitmap* GetActiveBitmap()`.  That function returned a pointer to one of `CBitmap m_bmFull`, `CBitmap m_bmHalf`, or `CBitmap m_bmSmall`.  In 47b05, for const- and null- correctness of Tile.h, I changed this to `CBitmap& GetActiveBitmap()`.  The function returned a reference to one of `OwnerPtr<CBitmap> m_bmFull`, `OwnerPtr<CBitmap> m_bmHalf`, or `CBitmap m_bmSmall` with the appropriate dereferencing to get to the actual CBitmap.  This allows the caller to replace the content of `m_bmFull`, `m_bmHalf`, or `m_bmSmall` as appropriate, just as the original version did.

3)  I have changed some declarations from `CBitmap` to `OwnerPtr<CBitmap>` because, as noted above, MFC does not support copy or move operations.  They do have `Attach()` and `Detach()` to manipulate ownership of the underlying GDI object, but I think the move notation (provided by `OwnerPtr<>`) is more attractive, and I am willing to accept with the mild overhead of the extra heap allocation.  (For comparison, note that wxWidgets's `wxBitmap` supports copy operations because they use reference counting, so copy can be implemented relatively cheaply.  I am only supporting move rather than copy because that's simpler both to implement and execute.)

4)  The overall result is that `GetActiveBitmap() = CloneBitmap(*m_pEditView->GetCurrentViewBitmap());` creates a new (independent) `CBitmap` that looks identical to `m_pEditView->GetCurrentViewBitmap()`, and stores that new `CBitmap` in one of `m_bmFull`, `m_bmHalf`, or `m_bmSmall`.  I believe that is the same net effect as the original code.

DLLarson: Ok... I get the idea of what you are doing. 

I just have a problem with the unusual syntax of a function being assigned a value. I just haven't seen that approach much in my C/C++ programming over the years.

I kinda agree with this statement: 

>The ability to return lvalues from functions looks pretty obscure, yet it is useful when you are doing advanced stuff like implementing some overloaded operators.

From this post:

https://www.internalpointers.com/post/understanding-meaning-lvalues-and-rvalues-c

I probably wouldn't have had such a reaction if I saw comment stating that the function returns an lvalue that may be used for assignments.

Anyhoo... I get the idea now.

-Dale

GShr/Board.cpp
```diff
@@ -186,16 +180,246 @@ CDrawList* CBoard::GetTopDrawing(BOOL bCreate)
...
+                    AfxThrowInvalidArgException();
+            }
+        default:
+            AfxThrowInvalidArgException();
+    }
+}
+
+/* recognize the Peter Lomax Ogre/GEV board style where "sticking
+    out" cells are empty, and rest of edge is filled */
+bool CBoard::IsGEVStyle(Edge e) const
```
DLLarson, line 261: Just curious... Is there a long-lasting link to a graphic that could be added to the comment that illustrates this case?

-Dale

DLLarson:
Ok... I'm done reviewing the code. I'm assuming a non-squash merge should be used.

A couple questions...

1) Are there test files that demonstrate the changes to the geomorphic stuff? Such as an Ogre/GEV gamebox etc?
2) How is the file version compatibility affected by this PR?

-Dale

wsu-cb:
> Ok... I'm done reviewing the code. I'm assuming a non-squash merge should be used.
> 
> A couple questions...
> 
> 1. Are there test files that demonstrate the changes to the geomorphic stuff? Such as an Ogre/GEV gamebox etc?

I'm pretty sure the Peter Lomax gbx that motivated the IsGEVStyle() concept is already in GamesLib-CBTestFiles.7z as _SJG\OgreGev2\Ogre_v10.zip.

> 2. How is the file version compatibility affected by this PR?
> 

During the cb64 work, I had already started digging into the geomorphic board code, so in the cb64 code, I added a reserved uint8_t to CGeoBoardElement to hold the board rotation info, and the cb64 code always set it to 0, which is the value for unrotated boards in the new geomorphic board code.  The square board support just requires holding the BoardID, which was already required for the hex boards anyway, it just permits additional values to be handled.  Likewise, supporting even-numbered hex dimensions also just adds support for additional values that the file could already hold.  The support for the Peter Lomax-style boards does change how such boards will display between the cb64 branch and the new branch, but I argue that the old display was far enough away from expected that I doubt anyone ever actually used geomorphic maps with that .gbx once they saw the result.  So all files that the cb64 branch properly handled will be written in the same format with the new geomorph code, and hence will still be understood by the cb64 branch.  However, if someone creates a geomorphic board using features that the cb64 branch doesn't provide, e.g., geoboards with square unit boards or rotated unit boards, I'm not too sure how much error detection the cb64 branch will do to reject that board cleanly.

Unfortunately, adding file storage space for rotated unit boards does mean that cb310 format readers won't understand either cb64 or geomorph branch files, but there is a file format version that the cb310 readers know to reject.

To be honest, I would have preferred to make the cb64 file version 4.0, and the geo version 4.1, to allow for clear recognition by older readers that the file was newer, but once we do that, then all intermediate versions should be separate.  So at that point, my current work on multi-sided-pieces would be version 4.2, and we would likely see significant version jumps between releases.  Also, if we are each working on the code, and each create file format changes, we're still stuck with incompatible files with the same file format version number, so we would still need a lot of cooperation to resolve file version problems.

> -Dale

GShr/Board.cpp
wsu-cb: Here's a screenshot:

image

So I think we can put a link to this screenshot in the comment now.  I assume github won't throw away this image, will it?

wsu-cb:
> So boiling this all down...If I run CB without the `/id32cb64` switch will the current CB in wide use (3.10) still work if these commits are absorbed into `master`? This is my primary concern as the switch was--in my understanding--how we would segregate the new additions until an official bump up to 4.x is released.
> 

If you look at CGeomorphicBoard::Serialize(), you will see that running CB-geomorph without `/id32cb64` will throw an exception when trying to save a geoboard that CB 3.10 won't handle properly.  Geoboards that satisfy CB 3.10 expectations can be saved by CB-geomorph (without `/id32cb64`) and used by CB 3.10.

The partial exception is that CB-geomorph will still save geoboards with the Peter Lomax IsGEVStyle(), but CB 3.10 will interpret them in the old, not-very-useful, way.  I dithered over whether to make CB-geomorph reject saving IsGEVStyle() geoboards without `/id32cb64`, but decided that, since CB 3,10 already saved them, CB-geomorph can follow suit.

You understand my intent correctly; the goal of running without `/id32cb64` is to ensure CB 3.10 can still read files written by new CB code.  However, that requires the user to limit himself to CB 3.10 features; making use of new CB features may prevent generating a CB 3.10 file.  Try to save often so you don't lose too much work due to being unable to save in the old format.

DLLarson:
Excellent!

Then I'm good to go to merge this if you're still good with it.

-Dale

wsu-cb:
> Then I'm good to go to merge this if you're still good with it.
> 
> -Dale

You imply in https://github.com/CyberBoardPBEM/cbwindows/pull/78#discussion_r830649279 that you would like the comment to include a link to the screenshot of an IsGEVStyle() board.  Let me know if you want me to adjust the comment,  If you are comfortable adding a link to the screenshot to the comment yourself, that's fine too.

DLLarson:
>ou imply in https://github.com/CyberBoardPBEM/cbwindows/pull/78#discussion_r830649279 that you would like the comment to include a link to the screenshot of an IsGEVStyle() board. Let me know if you want me to adjust the comment, If you are comfortable adding a link to the screenshot to the comment yourself, that's fine too.

I forgot about that. Even better, maybe a link to the PR might be even better. Either way may as well do it and be done.

Thanks,
Dale

Issue #14, opened 2020-09-13
pieces with arbitrary side count


wsu-cb:
It would be nice to allow pieces with more than two sides to handle things like Steve Jackson Games Ogre infantry (3 sides), or GMT Commands and Colors (often 4 sides).  It would also allow pieces with 6 sides to emulate the War of the Ring custom dice (although some other features would also be needed).

wsu-cb:
#81 supports pieces with up to 100 sides, so close this.

PR #81, opened 2022-04-06
Create many-sided pieces

wsu-cb:
This is the first step to support pieces with many sides (Enhancement #14).  It allows CBDesign to create and edit pieces with many sides.  The list boxes can show all the sides, but most of the rest of the system doesn't support the extra sides yet.

Warning:  the extra sides require a change in file format as compared to file version 4.0-geomorph.  However, as usual, I haven't changed the file format number as stored in the files.  Therefore, please be very careful about which files are loaded into which intermediate CB versions.  Also, I expect there will be more file format changes needed as I add support for many-sided pieces to other parts of CB.

Dale, please let me know what you think of this so far.


DLLarson:
Hi Bill,

I'll be looking this over in the next few days.

-Dale

GM/DlgPedt.cpp
```diff
@@ -37,8 +37,13 @@ static char THIS_FILE[] = __FILE__;
...
 
 
-CPieceEditDialog::CPieceEditDialog(CWnd* pParent /*=NULL*/)
-    : CDialog(CPieceEditDialog::IDD, pParent)
+CPieceEditDialog::CPieceEditDialog(CGamDoc& doc, PieceID pid, CWnd* pParent /*=NULL*/)
+    : CDialog(CPieceEditDialog::IDD, pParent),
+    m_pDoc(&doc),
+    m_pid(pid),
+    m_tbl(1, m_pid),
+    m_pTMgr(*m_pDoc->GetTileManager()),
```
DLLarson, line 45: Just strange looking that these near identically named members (`m_pTMgr` and `m_pPMgr`) used different looking levels of indirection.

wsu-cb: 1)  When I change the type of a variable, I generally try to avoid changing its name.  I think it makes it easier to understand what I readily admit are pretty extensive changes.
2)  I generally try to use the most restrictive type that still permits the code to compile.

So we wind up with inconsistencies like this for pre-existing variables.  I acknowledge that not changing the name now may create confusion in the future when someone looks at the replacement code without looking at the previous code.

If you want me to remove the `p` markers when I change something into a reference in the future, I can try to remember that, but I think it will make diffs harder to understand.  At the cost of making even larger pull requests, another option is to make two commits, one like the ones I currently make, and a follow-up commit that removes obsolete `p` marks.  I think that will keep the individual commits easier to understand, but obviously creates even more commits and larger pull requests.

DLLarson: Forget all that. It was just a gut reaction. It's fine as is.

There is much bigger fish to fry.

-Dale

GShr/Versions.h
```diff
@@ -96,18 +96,18 @@ inline int GetSaveFileVersion()
     static const int retval = [] {
         struct FileFlagParser : public CCommandLineInfo
         {
-            bool id32cb64 = false;
+            bool id32cb64sidemany = false;
             virtual void ParseParam(const char* pszParam, BOOL bFlag, BOOL bLast) override
             {
-                if (bFlag && strcmp(pszParam, "id32cb64") == 0)
+                if (bFlag && strcmp(pszParam, "id32cb64sidemany") == 0)
```
DLLarson, line 102: Question... Should I be specifying `/id32cb64sidemany` when I run the app for multifaced operation?

-Dale

wsu-cb: If you just want to create and edit many-sided pieces, you don't need the switch.  However, you will need it to save a file with many-sided pieces.  Presumably, you will typically want to save your work, so I would typically run with the switch.

(I changed /id32cb64 to /id32cb64sidemany as in indicator that the current file format is no longer compatible with the previous file format 4.0.  One-or-two-sided pieces can still be saved in CB3.1 format.)

DLLarson: Hi Bill,

I've had a chance to scan the code and I think you're starting in a good place. Obviously the most challenging thing is accounting for the multi-side stuff throughout the UI such that it operates in sane and intuitive fashion. I think it's wise to retain the concept of a top side as it is the anchor point of a piece. Expanding the single back side into multiples makes sense.

I was thinking of unusual use cases for this and modeling a multi-faced die having special images for each face came to mind. It would be cool to be able to pull a piece out of a tray and have it result in a random side selection. I'm sure people will think up all kinds of applications for multi-sided pieces that wander far from representing more piece information for instance. It should be interesting.

I say charge ahead!

I'm assuming that we shouldn't fold this into `master` yet. In fact we may want to create a global working branch so we can collect inspected PR's in one place without risking damaging the main code until it's ready. That would be easier to manage than a constantly expanding a single PR.

-Dale

wsu-cb: > Hi Bill,
> 
> I've had a chance to scan the code and I think you're starting in a good place. Obviously the most challenging thing is accounting for the multi-side stuff throughout the UI such that it operates in sane and intuitive fashion. I think it's wise to retain the concept of a top side as it is the anchor point of a piece. Expanding the single back side into multiples makes sense.
> 
> I was thinking of unusual use cases for this and modeling a multi-faced die having special images for each face came to mind. It would be cool to be able to pull a piece out of a tray and have it result in a random side selection. I'm sure people will think up all kinds of applications for multi-sided pieces that wander far from representing more piece information for instance. It should be interesting.
> 

As I said in Issue #14, the non-standard War of the Ring dice were one of my specific motivations for this work.

I certainly think it would make sense to give the trays the option to put a piece on a random side when pulled from the tray.  However, I plan to put a context menu on pieces in the selected-piece window that will allow choosing a random side (for dice), as well as flipping the piece to the side where the user brought up the menu (for strength-point style pieces).

> I say charge ahead!
> 
> I'm assuming that we shouldn't fold this into `master` yet. In fact we may want to create a global working branch so we can collect inspected PR's in one place without risking damaging the main code until it's ready. That would be easier to manage than a constantly expanding a single PR.
> 
> -Dale

I definitely don't want this merged to master any time soon.  I posted it now because this is so different from the relatively minor (in terms of concept, not lines of code) changes I had been making in previous PRs.  I wanted to be sure you approved the general direction I am going.

I think a feature branch would make a lot of sense for this.  However, I am likely to make a lot of force pushes to the branch.  In my previous PRs, I had always kept my branches local and done a lot of rebase clean-up work before pushing for review.  If I push to github earlier, those rebases won't be just on my local repo.  There is the middle ground of keeping the feature branch only in my fork, and doing the rebases there, and only sending a PR to the main CB repo when I am ready for your review.  What's your preference?

DLLarson:
>There is the middle ground of keeping the feature branch only in my fork, and doing the rebases there, and only sending a PR to the main CB repo when I am ready for your review.

I think this approach would be fine. You'll have your local copy and a backup in your fork just-in-case. When ready just post to the PR.

I'm just trying to avoid reinspection of code I've already looked at. I understand that this is somewhat tough to avoid since the same source code files will evolve along with the additional UI (etc...) work so they will show up again. But the adding new features should definitely trump my discomfort with redundancy. ;)

-Dale

DLLarson (changes requested):
Whew! That was a heap 'o code to get through!

I'd like to run the code against some files. Have you cooked up any test files for these changes?
It would save me some time trying out the features.

-Dale

GP/PalTray.cpp
```diff
@@ -909,75 +960,152 @@ void CTrayPalette::OnUpdateEditElementText(CCmdUI* pCmdUI)
...
+BOOL CTrayPalette::OnActTurnOver(UINT id)
 {
     CArray<int, int> tblListSel;
     int nNumSelected = m_listTray.GetSelCount();
     tblListSel.SetSize(nNumSelected);
     m_listTray.GetSelItems(nNumSelected, tblListSel.GetData());
 
-    m_pDoc->AssignNewMoveGroup();
-    for (int i = 0; i < tblListSel.GetSize(); i++)
+    CArray<int, int> tblListSubjects;
```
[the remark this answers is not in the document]
wsu-cb: I think it was because tblListSubjects has to have the same type as tblListSel, and there was no specific need to change tblListSel.

GShr/CyberBoard.h
```diff
@@ -974,23 +1034,147 @@ class XxxxIDTable
     size_t      m_nTblSize;     // Number of alloc'ed ents in tile table
 };
 
+template<typename KEY, typename ELEMENT,
+    size_t baseSize, size_t incrSize,
+    bool saveInGPlay>
+class XxxxIDTable<KEY, ELEMENT, baseSize, incrSize, saveInGPlay, false>
+{
```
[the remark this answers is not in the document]
wsu-cb: There was quite a bit of v3.1 code that was the same between `CMarkManager`, `CPieceManager`, and `CTileManager`, so I consolidated that code into a template.  At the time, I thought the `Serialize()` code for `CTileManager` differed from `CMarkManager` and `CPieceManager` in using #ifdef to skip the save of the ID table in GPlay, so I used this template argument to address that distinction.  However, in reviewing the code today to write this comment, I now think both the v3.1 code and the new code use #ifdef in a way that means XxxxID<TileID>::Serialize() is never even called when saving in GPlay, and that this argument is therefore unnecessary.  Whoops.

GShr/LBoxMark.cpp
```diff
@@ -88,23 +89,34 @@ GameElement CMarkListBox::OnGetHitItemCodeAtPoint(CPoint point, CRect& rct)
...
+    }
+    else
+    {
+        return Invalid_v<GameElement>;
+    }
 }
 
 void CMarkListBox::OnGetTipTextForItemCode(GameElement nItemCode,
-    CString& strTip, CString& strTitle)
+    CString& strTip, CString& strTitle) const
```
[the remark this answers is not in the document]
wsu-cb: It looks to me like you were allowing for the possibility of using `TTM_SETTITLE`, but that no code ever actually took advantage of that option.

DLLarson (approved):

DLLarson:
I approved the code changes but I'm pretty sure you don't want them merged to master yet.

Am I correct?

-Dale

wsu-cb:
> I approved the code changes but I'm pretty sure you don't want them merged to master yet.
> 
> Am I correct?
> 
> -Dale

Actually, I think this PR is ready for merge.  I have done some testing of it, and really serious testing will probably be done only after I create `.gam`s for Space Empires that will also make use of #86.  Unfortunately, building the `Space Empires.gbx` isn't a quick process.

DLLarson:
>Actually, I think this PR is ready for merge. I have done some testing of it, and really serious testing will probably be done only after I create .gams for Space Empires that will also make use of https://github.com/CyberBoardPBEM/cbwindows/issues/86.

Ok! To master it is.

An exciting moment!

-Dale

PR #82, opened 2022-04-15
fix newly created Piece objects having an owner

wsu-cb:
CB3.1 used GlobalAlloc(GHND) to allocate CPieceTable memory,
which guarantees that allocated memory is zero-initialized.
Therefore, newly created Piece objects had no owner even though
Piece::SetUnused() didn't initialize m_dwOwnerMask.  Now that
malloc() is allocating memory, this is no longer true.
Therefore, make all of the XxxxIDTable<> initialize functions
explicitly initialize all member variables.

### DIFF
--- a/GP/PPieces.cpp
+++ b/GP/PPieces.cpp
@@ -552,6 +552,7 @@ void Piece::SetUnused()
 {
     m_nSide = uint8_t(0xFF);
     m_nFacing = uint16_t(0);
+    m_dwOwnerMask = uint32_t(0);
 }
 
 
--- a/GShr/Marks.h
+++ b/GShr/Marks.h
@@ -64,7 +64,7 @@ struct MarkDef
     enum { flagPromptText = 0x8000 };
 
     // -------- //
-    void SetEmpty() { m_tid = nullTid; }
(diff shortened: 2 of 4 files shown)


DLLarson:
`calloc` could've been used as a substitute for the prior Win32 GlobalAlloc.

-Dale

DLLarson (approved):
Looks good to me. This is more explicit than counting on the memory allocator to zero the memory block.

Is this something that can be merged into master or an addition to your working branch?

-Dale

wsu-cb:
> Looks good to me. This is more explicit than counting on the memory allocator to zero the memory block.
> 

That's why I chose this approach instead of `calloc`.

> Is this something that can be merged into master or an addition to your working branch?
> 

Yes, I discovered and fixed this in my working branch initially, but I cherry-picked this to master for this PR.   I don't want to wait for my working branch to be merged because this causes invalid files to be created, and I don't want anyone else to get hurt by that.

> -Dale

DLLarson:
> I don't want to wait for my working branch to be merged because this causes invalid files to be created, and I don't want anyone else to get hurt by that.

Thanks for the clarification. Merging...

-Dale

PR #87, opened 2022-09-04
Owned board hidden movement

### LINKED ISSUE #86: fully hidden pieces on owned board
Currently, a non-owner of a board can still see pieces on an owned board; the only limitation is the general restriction that a non-owner only sees the front of owned pieces.  However, for some games, it would be useful to completely hide pieces on an owned board.  For example, fully hidden movement, or an empire's economics record sheet.

I am thinking that it would be reasonable for a board with CPlayBoard::m_bNonOwnerAccess set to false to prevent showing pieces at all.

wsu-cb:
Resolve #86


wsu-cb:
Drat.  After more testing, I would like to change the playback display when flipping owned pieces.  Therefore, I have converted this PR to a draft.

GP/GamDoc.cpp
```diff
@@ -1410,7 +1410,7 @@ void CGamDoc::OnPbckPrevious()
         // search for the final visible move index.
         m_astrMsgHist.RemoveAll();
 
-        while ((m_nCurMove = m_pMoves->DoMove(*this, m_nCurMove, FALSE)) < nPrvMove &&
+        while ((m_nCurMove = m_pMoves->DoMove(*this, m_nCurMove)) < nPrvMove &&
```
wsu-cb, line 1413: I'm somewhat concerned about this change.  I don't understand what situation the original code's `FALSE` was handling, so I don't know whether this change could be causing a problem.

GP/GamDoc3.cpp
```diff
@@ -133,7 +133,7 @@ void CGamDoc::SerializeMoveSet(CArchive& ar, CHistRecord*& pHist)
         ar >> verMajor;
         ar >> verMinor;
         if (NumVersion(verMajor, verMinor) >
-            NumVersion(fileGsnVerMajor, fileGsnVerMinor) &&
```
DLLarson, line 136: Why the switch from `...Gsn...` to `...Gmv...`?
Was this a bug?

GP/GameBox.cpp
```diff
@@ -114,7 +114,7 @@ BOOL CGameBox::Load(CGamDoc* pDoc, LPCSTR pszPathName, CString& strErr,
         ar >> verMajor;
         ar >> verMinor;
         if (NumVersion(verMajor, verMinor) >
-            NumVersion(fileGsnVerMajor, fileGsnVerMinor) &&
+            NumVersion(fileGbxVerMajor, fileGbxVerMinor) &&
```
DLLarson, line 117: Same question about why this changed. Bug?

GShr/LBoxGfx2.cpp
```diff
@@ -116,7 +116,7 @@ void CGrafixListBox2::UpdateList(BOOL bKeepPosition /* = TRUE */)
     SetRedraw(TRUE);
     if (bKeepPosition)
     {
-        SetScrollPos(SB_HORZ, horzScroll);
+        SendMessage(WM_HSCROLL, MAKELONG(int16_t(SB_THUMBPOSITION), value_preserving_cast<int16_t>(horzScroll)), NULL);
```
DLLarson, line 119: Was this for a bug?

GShr/LBoxGrfx.h
```diff
@@ -300,7 +300,7 @@ class CGrafixListBoxData : public BASE_WND
         SetRedraw(TRUE);
         if (bKeepPosition)
         {
-            SetScrollPos(SB_HORZ, horzScroll);
+            SendMessage(WM_HSCROLL, MAKELONG(int16_t(SB_THUMBPOSITION), value_preserving_cast<int16_t>(horzScroll)), NULL);
```
DLLarson, line 303: Was this for a bug?

GShr/Versions.h
```diff
@@ -99,18 +102,20 @@ inline int GetSaveFileVersion()
     static const int retval = [] {
         struct FileFlagParser : public CCommandLineInfo
         {
-            bool id32cb64sidemany = false;
+            int version = NumVersion(4, 0);
             virtual void ParseParam(const char* pszParam, BOOL bFlag, BOOL bLast) override
             {
-                if (bFlag && strcmp(pszParam, "id32cb64sidemany") == 0)
+                static const std::regex re(R"(filever:([[:digit:]]+)\.([[:digit:]]+))");
```
DLLarson, line 108: My first look at the `regex` support. Nice!

DLLarson (changes requested):
I just had some questions.

GP/GamDoc.cpp
DLLarson: I'm not clear on the issue here.

GShr/LBoxGfx2.cpp
wsu-cb: Yes.  See https://github.com/CyberBoardPBEM/cbwindows/pull/87/commits/de96b8333c9cd5a71fbb1c8e343c55dfc3d5fd46.  (`SetScrollPos()` moved the scroll thumb without actually changing the visible portion of the window!)

GShr/LBoxGrfx.h
wsu-cb: See https://github.com/CyberBoardPBEM/cbwindows/pull/87#discussion_r975920329.

GP/GameBox.cpp
wsu-cb: This code is reading a `.gbx`, so I do think it should be checking against `fileGbxVerMajor`/`fileGbxVerMinor`.

GP/GamDoc3.cpp
wsu-cb: This code is reading a `.gmv`, so I think checking `fileGmvVerMajor`/`fileGmvVerMinor` is more appropriate.

GP/GamDoc.cpp
wsu-cb: I'm sorry; I can't tell if you mean you don't understand my comment, or if you mean you don't remember why the original code used `FALSE`.

GShr/LBoxGfx2.cpp
DLLarson: Dang! Nice find!

wsu-cb:
Dale, are you still reviewing this?

DLLarson:
I just had some questions in the review before merging (see above about back stepping. That's why I didn't pull the trigger on the merge.

-Dale

wsu-cb:
> I just had some comments in the review before merging (see above about back stepping) so I wanted to see if you were good with them. That's why I didn't pull the trigger on the merge.

Did you see https://github.com/CyberBoardPBEM/cbwindows/pull/87#discussion_r975936237?

DLLarson:
> > I just had some comments in the review before merging (see above about back stepping) so I wanted to see if you were good with them. That's why I didn't pull the trigger on the merge.
> 
> Did you see [#87 (comment)](https://github.com/CyberBoardPBEM/cbwindows/pull/87#discussion_r975936237)?

Yes and I replied with this:

>so this code is the move back a step processing. It goes back to one move before the move we're stepping back to so it can replay it to get to the desired state. This code is tricky and my memory is definitely fuzzy here but this comment:

        // OK...we know there is at least one move to do PRIOR to the move
        // we are stepping up to. Find it so we can turn off silent mode for
        // that move.
>seems to suggest that the argument bAutoStepHiddenMove value should be overridden with FALSE so that move is visible.

So the code changes seems to reverse what the code did before (by default initialization) and I feel this may be wrong so--as a minimum-- I expected you to react to that as it seems wrong.

-Dale

DLLarson (approved):

DLLarson (approved):
Looks good.

PR #90, opened 2022-10-22
more ownership / CPieceObj::m_rctExtent interaction work

wsu-cb:
The problem of CPieceObj::m_rctExtent being different for different players also shows up in playback, not just .gmv validation.  So whenever the CGamDoc's main CPieceTable Serialize() or Restore() run, force a recalculation of the CGamDoc's CDrawLists.

### DIFF
--- a/GP/GamState.cpp
+++ b/GP/GamState.cpp
@@ -57,7 +57,13 @@ BOOL CGameState::CompareState(const CGamDoc& doc) const
     if (!doc.GetPieceTable()->Compare(*m_pPTbl))
         return FALSE;
     // Compare the play boards
-    if (!doc.GetPBoardManager()->Compare(*m_pPBMgr))
+    /* compare must be relative to the doc's piece table, not
+        game state's, because ownership might be different,
+        which can cause CPieceObj::m_rctExtent to differ */
+    // const_cast safe because we're just using temp for compare
+    OwnerPtr<CPBoardManager> temp = m_pPBMgr->Clone(const_cast<CGamDoc&>(doc));
+    temp->OnPieceTableChanged();
+    if (!doc.GetPBoardManager()->Compare(*temp))
         return FALSE;
     // Compare the trays
     if (!doc.GetTrayManager()->Compare(m_pYMgr))
--- a/GP/PBoard.cpp
+++ b/GP/PBoard.cpp
@@ -506,6 +506,13 @@ void CPlayBoard::Serialize(CArchive& ar)
     }
 }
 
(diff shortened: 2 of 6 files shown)


wsu-cb:
This pull request fixes a bug as described in the commit comment (or at least I claim it does).  However, I am not marking this ready for review because I think this fix is somewhat ugly.  I want to propose a different, perhaps somewhat radical, approach, and see how you feel about it.

I think the fundamental problem is that for most of the `CDrawObj` subclasses, the `m_rctExtent` is solely determined by operations directly performed on the `CDrawObj` itself, but that is not always true for `CPieceObj`, since ownership of a `PieceID` can be changed by starting playback of a move.  (At least, given that I definitely want to allow different sides to have different sizes, and therefore ownership affects `m_rctExtent`.)  Since there are "remote" factors affecting `CPieceObj`'s `m_rctExtent`, keeping `m_rctExtent` up-to-date requires that these other "remote" factors notify `CPieceObj` of the new state.

I think it would be cleaner if `CPieceObj` did not actually store `m_rctExtent` at all, but rather calculated it when needed.  I am therefore proposing that there are two different subclasses of `CDrawObj`, one `CPieceObj` with no `m_rctExtent` member, and one that would be a base class for all the other existing `CDrawObj` subclasses that does contain an `m_rctExtent` member.  (`CDrawObj` would need some additional changes to make existing uses of `m_rctExtent` into virtual function calls instead.)  For `CPieceObj`, I think it should maintain a `m_pntCenter`, since the existing `CPieceObj::ResyncExtentRect()` treats the center as the fixed point of the object, and then get size info from `m_pid` when the momentary value of `m_rctExtent` is required.  I realize this would probably be a performance pessimization, but I think it would lessen the coupling between `CPieceTable` and a number of other classes that this pull request introduces.

DLLarson:
>However, I am not marking this ready for review because I think this fix is somewhat ugly.

Understood.

>I think it would be cleaner if CPieceObj did not actually store m_rctExtent at all, but rather calculated it when needed. ... I realize this would probably be a performance pessimization, but I think it would lessen the coupling between CPieceTable and a number of other classes that this pull request introduces.

I think the additional computational load would be nearly nothing in the greater scheme of things. Many of the efficiencies the code does harkens back to getting a Win16 platform to perform well. The world has moved far beyond those CPU limitations so I say go for it. Plus... I'm always interested in reducing coupling and increasing cohesion.

-Dale

wsu-cb:
superseded by #91

PR #95, opened 2022-11-09
restrict drag-drop destination to board limits

### LINKED ISSUE #92: drag-drop can put objects far off board
[big-and-small-boards.zip](https://github.com/CyberBoardPBEM/cbwindows/files/9947011/big-and-small-boards.zip)

The attached .zip contains a `.gbx` and `.gsn` that shows that it is possible for drag-and-drop to get put the dropped objects far off the board.  To see the problem, follow these steps:

1)  Open the `.gsn`
2)  Rubber-band-select the two markers in the bottom-right corner of the big board.
3)  Left-click and hold the right-hand marker of the pair of markers to start dragging.
4)  Drag the right-hand marker to the top-left cell of the small board.  Before dropping the marker, notice that there is already something wrong happening since the drag-tracker squares are not displayed.
5)  Drop the right-hand marker in the top-left cell of the small board.  Note that the `CSelectedPieceView` shows the markers, but the `CPlayBoardView` does not.
6.  View the small board at small scale.  You can probably find the two gray squares for the markers near the top of the view, but far to the right of where the board ends.

wsu-cb:
Fix #92

### DIFF
--- a/GP/VwPbrd.cpp
+++ b/GP/VwPbrd.cpp
@@ -760,8 +760,9 @@ LRESULT CPlayBoardView::DoDragSelectList(WPARAM wParam, DragInfo& pdi)
         BOOL bXOK, bYOK;
         if (!IsRectFullyOnBoard(rctObjs, &bXOK, &bYOK))
         {
-            sizeDelta.cx = bXOK ? sizeDelta.cx : 0;
-            sizeDelta.cy = bYOK ? sizeDelta.cy : 0;
+            CRect temp = rctObjs;
+            LimitRect(temp);
+            sizeDelta += temp.TopLeft() - rctObjs.TopLeft();
         }
         if (sizeDelta.cx != 0 || sizeDelta.cy != 0) // Check 'em again (what a pain!)
             pSLst->Offset((CPoint)sizeDelta);


wsu-cb:
Note that this fixes the most glaring problem, but there are still some problems this doesn't fix:

1)  The two markers could be so far apart that both can't be on the destination board at the same time.
2)  A single object could be so large it can't fit on the destination board.
3)  Drag the markers from the big board to the small board, but do NOT actually drop the markers there.  Continue dragging until the mouse cursor is in the big board's `CSelectPieceView`.  The mouse cursor changes to show that dropping there is forbidden, but release the mouse anyway.  This seems to put CB in a strange state.  The `CSelectPieceView` shows the markers as selected, but the board does not (re)draw the markers with handles.  Also, the markers can still be dragged despite not showing handles, but if they are dragged, the tracking rectangles don't properly follow the mouse cursor.  If I switch the big board to its tiny view, I can see that dragging the markers shows the tracking rectangles for the markers far away from the mouse cursor.  So apparently drag-drop doesn't properly clean up a forbidden drop.

Nevertheless, I would still suggest applying this change to fix the originally-discovered problem.

DLLarson:
It's definitely behaving better than before. But is some cases I get artifacts. When I drag the mouse into the small board all worked well but when I when back to the large board and dragged along the right edge I see this:

image

But it doesn't always happen. It's almost like a race condition. I was running both the 64bit and 32bit debug versions of this PR to test.

Regarding your other points.

1. In the drag rectangle too large case the best approach may be to shrink the box until it fits along the top, left, or both (if too large in both dimensions) and proportionately relocate the dragged items.
2. In the dragged item too large it's tricky. One could just butt it up to the left/top edge and render the item it a manner that get clipped. Another approach would be to change the drag target cursor to something the indicates the target location can't accept the drop.
3. I couldn't reproduce this for some reason. I wonder if it's because I'm running on a 4K monitor. If you could give me more clarity on this: "Continue dragging until the mouse cursor is in the big board's CSelectPieceView."

-Dale

DLLarson (commented):
I'm holding the PR until you see my comments.

wsu-cb:
Regarding 1. and 2., I was leaning toward rejecting the drop in both cases, which I think is what you are saying as well.

Regarding 3., I don't have any trouble reproducing it.  Indeed, I now see that I can get into this problem without ever viewing the small board.
1) Start with the original .gsn.
2) Select just the rightmost marker.
3) Drag the marker approximately two big-board cells to the right, so that the mouse is in the big board's tiny board window, and drop the marker in this forbidden window.
4) The result is that the marker stays displayed in its original location, and the `CSelectPieceView` shows the marker still selected, but the marker is displayed with no handles.  Also, if I drag the marker again, the tracking rectangle is in the wrong place.  Note that to see the tracking rectangle located incorrectly, the second drag has to be performed without unselecting and then reselecting the marker.

Regarding 4., which is what I will call your drawing artifacts, I can reproduce the problem as well.  And, again, I can do so without ever viewing the small board.
1) Start with the original .gsn.
2) Select just the rightmost marker.
3) Drag the marker to the right, but don't actually move the cursor into the tiny board view.  Instead move the mouse cursor just far enough to the right to have the cursor just slightly to the left of the scroll bar.  I see the tracking rectangle start blinking.  Now start moving the mouse cursor up and down.  The blinking tracking rectangle will leave artifacts, which actually sort of make sense, because the XOR drawing logic could easily miss its tracking rectangle erase if the mouse moves at a bad time.

wsu-cb:
> Regarding 1. and 2., I was leaning toward rejecting the drop in both cases, which I think is what you are saying as well.
> 

Correction:  what you were saying for 2., but not 1.

DLLarson:
>Regarding 1. and 2., I was leaning toward rejecting the drop in both cases, which I think is what you are saying as well.

I think that would be fine as well for both cases. It's consistent.

>... 4. The result is that the marker stays displayed in its original location, 

Ok. I can see what you're talking about now. Thanks for the alternate repo procedure. Of course this is a bug.

>Regarding 4., which is what I will call your drawing artifacts, .... because the XOR drawing logic could easily miss its tracking rectangle erase if the mouse moves at a bad time.

This is what it made it feel like a race condition to me. This artifact is in the shipping version as well. This only seems to happen when dragging along a view edge. It doesn't happen when dragging along an edge when the auto-scroll is occurring which is more complicated to process. Weird.

This PR fixes some items so we should probably merge this change (as you originally requested) but the other issues should be entered as a new defect. I'll do the merge with your thumbs up. I just wonder why this hasn't been reported before since it's in the shipping bits. There's a sequence problem somewhere in the code that allows this to happen--perhaps a recursive code call without lock to prevent reentry into the same code.

-Dale

wsu-cb:
> This PR fixes some items so we should probably merge this change (as you originally requested) but the other issues should be entered as a new defect. I'll do the merge with your thumbs up. I just wonder why this hasn't been reported before since it's in the shipping bits. There's a sequence problem somewhere in the code that allows this to happen--perhaps a recursive code call without lock to prevent reentry into the same code.

Yes, I think this should be merged and three separate new bugs entered for 1./2., 3., and 4..

DLLarson:
I'll add the new issue items later today-ish.
-Dale

Issue #96, opened 2022-11-17
Handle Drag/Drop of selected objects into incompatible target views


DLLarson:
(Issues extracted from PR https://github.com/CyberBoardPBEM/cbwindows/pull/95)

Handle the following degenerative cases:

1. Selected objects could be so far apart that both can't be contained on the target view at the same time without changing their layout.
2. A single object could be so large it can't fit on the destination view.

DLLarson:
The desired solution is to indicate the invalid drop locations using a new Cursor graphic that uniquely identifies this problem to the user.

-Dale

DLLarson:
We need to reopen this issue.

I discovered that the ability to drag tiles onto playing board in `CBDesign.exe` is broken. No matter where I drag the tiles the "drop-not-allowed" cursor is displayed.

-Dale

wsu-cb:
> I discovered that the ability to drag tiles from the Tile palette onto playing board in `CBDesign.exe` is broken. No matter where I drag the tiles the "drop-not-allowed" cursor is displayed.
> 
> -Dale

I can reproduce this.  I will investigate...

Issue #42, opened 2021-03-06
WinState.* appears to be out of date with respect to new MDI


wsu-cb:
It looks to me like the WinState.* code needs to be updated to account for the newer MDI style.  When I have two tab groups displayed (allowing me to drag pieces between boards), the WinState.* code doesn't seem to save/restore the split information.  That is, after I close and reopen the .gam, I am back to a single tab group.

DLLarson:
I'm wondering if the current "modernized" MFC is already performing this function and this older approach is fighting with the new MFC.

If so, the whole custom thing could be scrapped.

-Dale

PR #99, opened 2022-11-19
Drag drop fixes

wsu-cb:


GP/Cbplay32.vcxproj.filters
```diff
@@ -680,6 +680,9 @@
     <None Include="res\NoDrop.cur">
       <Filter>Resource Files</Filter>
     </None>
+    <None Include="res\NoDropTooBig.cur">
```
wsu-cb, line 683: I can't find a way post a comment on the actual .cur file, so I am commenting here:  The .cur file is a slightly modified version of the NoDrop.cur file to satisfy the requirement to have a special cursor for the too-big select list case.  However, it is not a good-looking cursor.  I'm hoping someone else will improve it.

GP/ToolPlay.cpp
```diff
@@ -629,6 +629,16 @@ void CPSelectTool::DoDragDropEnd(CPlayBoardView* pView, CPoint pntClient)
...
+    // if drop failed
+    if (pWnd != pView && !pView->GetSelectList()->empty())
+    {
+        // failed drop messed up selection list, so rebuild it
+        std::vector<CB::not_null<CDrawObj*>> listObjs;
+        CSelList& selList = CheckedDeref(pView->GetSelectList());
+        selList.LoadTableWithObjectPtrs(listObjs, CSelList::otAll, FALSE);
+        selList.PurgeList(TRUE);          // Purge former selections
+        pView->SelectAllObjectsInTable(listObjs);  // Reselect on this board.
+    }
```
wsu-cb, line 641: This code is an approximately minimal change to handle the case of a drop on a forbidden window, but I dislike it.  This is why I have marked this PR as draft.
1)  We don't have the `WM_DRAGDROP` message explicitly report success/failure, so I need to infer failure using `if (pWnd != pView && !pView->GetSelectList()->empty())`.  If WM_DRAGDROP recipients reported success, then we wouldn't need this indirect test.  Unfortunately, there is the case of a drop on a Window that doesn't even handle the message, and hence won't provide an explicit return value.  https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendmessage does not specify the return value when the message is not handled by the recipient, but experiments suggest the return value is 0.  As long as we can really depend on that, if we made all successful `WM_DRAGDROP` receivers return 1, we would change a lot more code, but I think we would be replacing this sneaky code coupling with explicit code coupling using return values.
2)  The clean-up logic really seems like it belongs in the calling view, but this function has no return code, so the caller can't know it failed.  I think making this function simply report the proposed `WM_DRAGDROP` success/fail return value to this function's caller would reduce code coupling.

Do these broader changes sound reasonable?

GP/Cbplay32.vcxproj.filters
DLLarson: I worked on the cursor image some. See if you like what a I attached to this post. If you do just replace the one in your project (I didn't want to add a commit to your PR.)

(Note that Github won't let me put the cursor file directly in the post hence the zip.)

[NoDropTooBig.zip](https://github.com/CyberBoardPBEM/cbwindows/files/10047647/NoDropTooBig.zip)

-Dale

GP/ToolPlay.cpp
DLLarson: Hi Bill,

>does not specify the return value when the message is not handled by the recipient, but experiments suggest the return value is 0

I believe that is the default for any message that isn't handled explicitly by the app's windows handler. The default windows hander routine just returns zero.

Since the message `WM_DRAGDROP` is custom to CB you can change it however you like. Perhaps returning values like 0=not handled, -1=failed, 1=succeeded for example.

>Do these broader changes sound reasonable?

I'm fine with making changes in this area. It's better than a maintainer needed to figure out how an inferred result was synthesized.

I've attached an updated version of your test board and scenario to add a huge tile as a marker to test the drop of an object that is too large. It works fine when dragging from board to board but doesn't work when dragging from a marker tray. Then it drops at the upper left corner and then graphically clips to the board boundaries. I suspect this would also happen when dragging pieces from trays as well.

-Dale

[Updated-big-and-small-boards.zip](https://github.com/CyberBoardPBEM/cbwindows/files/10047661/Updated-big-and-small-boards.zip)

GP/Cbplay32.vcxproj.filters
wsu-cb: That's much nicer.

DLLarson: >That's much nicer.

Great!

It took a bit of time to figure out VS2019's bitmap editor. I wish I could zoom in more that 8 times on a 4K screen. Their color picker was really small as well. I wonder if the more up-to-date VS community plays better on 4K screens.

-Dale

GP/ToolPlay.cpp
wsu-cb: > I've attached an updated version of your test board and scenario to add a huge tile as a marker to test the drop of an object that is too large. It works fine when dragging from board to board but doesn't work when dragging from a marker tray. Then it drops at the upper left corner and then graphically clips to the board boundaries. I suspect this would also happen when dragging pieces from trays as well.
> 

This is a really good point.  The problem is, what do we do about random markers?  Since the size of the markers can vary in the same tray, how do we know what size to use?  I think we have to assume the max marker width and max marker height.

DLLarson: >I think we have to assume the max marker width and max marker height.

If by "max" you mean the max size among all candidate objects that will be randomly selected from, I agree. That's the sensible thing to do. I think that it would be a rare situation but it still needs to be considered.

wsu-cb:
I think this PR is ready to merge.

GP/ToolPlay.cpp
```diff
@@ -97,11 +97,12 @@ void CPlayTool::OnMouseMove(CPlayBoardView* pView, UINT nFlags, CPoint point)
     SetCursor(AfxGetApp()->LoadStandardCursor(IDC_ARROW));
 }
 
-void CPlayTool::OnLButtonUp(CPlayBoardView* pView, UINT, CPoint point)
+bool CPlayTool::OnLButtonUp(CPlayBoardView* pView, UINT, CPoint point)
 {
     if (CWnd::GetCapture() != pView)
-        return;
+        return false;
```
DLLarson, line 103: This looks like you found a bug since the docs say that zero should be returned if the program processes it and the handler didn't return anything. 

https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-lbuttonup

Am I missing something here? I'm wondering why it didn't cause any obvious issues.

GShr/LBoxPiec.cpp
```diff
@@ -168,8 +168,10 @@ void CPieceListBox::OnItemDraw(CDC& pDC, size_t nIndex, UINT nAction, UINT nStat
 
 BOOL CPieceListBox::OnDragSetup(DragInfo& pDI) const
 {
-    pDI.m_dragType = DRAG_PIECE;
+    ASSERT(!"untested code");
```
DLLarson, line 171: Hmmmmm....Interesting!

DLLarson (commented):
I still need to test the build but I've gone through the code.

DLLarson:
I'm trying to drag around the marker trays to dock them and the UI seems busted. On my system it's really difficult to dock the tray on the sides of the windows. So far I've tried the 64bit version of CBPlay.

GM/Gm.cpp
```diff
@@ -53,7 +53,7 @@ static char szSectDisableHtmlHelp[] = "DisableHtmlHelp";
 
 /////////////////////////////////////////////////////////////////////////////
 
-UINT WM_DRAGDROP = RegisterWindowMessage("msgDragDrop");
```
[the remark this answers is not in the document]
wsu-cb: I wrote in the commit comment that, based on the pointer use of the WM_DRAGDROP message params, WM_DRAGDROP can only be used within its own process, not even to another instance of CB.  Therefore, I didn't think a unique message id mattered.  However, now that you have asked about this, I have thought some more, and now think that over-simplified the situation.
1)  Even though WM_DRAGDROP can only be SUCCESSFULLY used within its own process, it is still possible for the user to drag from one process to another, even a non-CB process, so I think you are correct to require system uniqueness.
2)  Unfortunately, I now think the problem is even worse.  WM_DRAGDROP passes pointers in its message params.  But if one CBPlay drags to another CBPlay, the pointer in the message param will be read in the DESTINATION process's address space, even though the message's actual data is in the SOURCE process's address space!  Best case, we get a crash due to the pointer being invalid.  Even worse, the pointer could be valid, but contain irrelevant data, leading to random behavior.  Can you see an error in my thinking here?

Assuming you agree with me, then the message params can't use pointers, or else the messages can't be passed across processes.  Since it's impossible to successfully drag pieces across processes, and it's really convenient for this code to use pointers within its own process, I think the simplest way to fix this problem is to make every process instance use a distinct message id.  So I now think the code should be
```
const UINT WM_DRAGDROP = RegisterWindowMessage(CB::Sprintf("msgDragDrop:%" PRIu32, GetProcessId(GetCurrentProcess())).c_str());
```
This will consume a message id per process, but I don't expect a lot of simultaneous CB processes in practice.

Opinions?

GP/Gp.cpp
```diff
@@ -61,7 +61,7 @@ static char szSectDisableHtmlHelp[] = "DisableHtmlHelp";
 // there isn't any common drag types. But just to be on the safe
 // side....
 
-UINT NEAR WM_DRAGDROP = RegisterWindowMessage("msgDragDropPlay");
```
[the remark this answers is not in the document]
wsu-cb: Ditto above answer

GP/ToolPlay.cpp
wsu-cb: This is MFC, not bare-metal Windows.  https://learn.microsoft.com/en-us/cpp/mfc/reference/wm-message-handlers-l-m?view=msvc-170 shows that most MFC handlers return void.  Also, this particular function isn't actually a direct message handler itself; it's called from a message handler.  So the return type of this function only matters to the actual message handler in the view.

wsu-cb:
I suspect the problem across save/load is a duplicate of #42.

I don't see any clear problems docking/undocking the piece/marker palettes otherwise.  On the other hand, I do think the docking behavior takes some experimenting to get a particular desired layout.

GM/Gm.cpp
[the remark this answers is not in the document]
DLLarson: Hi Bill,

There are many great points here. I believe your technique would work as each process message would be assigned a unique "atom" by the WinAPI. I did some more investigating after I read what you wrote and found this:

https://stackoverflow.com/questions/4406909/using-wm-user-wm-app-or-registerwindowmessage
https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-app

So it appears that for intra-app messages that are not used to implement windows controls should use WM_APP. I don't think WM_APP was a thing when I originally created CB and it's my first time bumping into it. This will stop any confusion with control based messaging but still doesn't guarantee non-collision with other apps--it just makes it less likely.

I think your idea of adding the processID to the message name is a winner! It completely solves problem.

-Dale

GShr/LBoxPiec.cpp
DLLarson: >Assuming you agree with me, then the message params can't use pointers, or else the messages can't be passed across processes. 

Just an FYI.... The Windows message `WM_COPYDATA` is used to pass data between processes.

-Dale

DLLarson:
>I don't see any clear problems docking/undocking the piece/marker palettes otherwise. On the other hand, I do think the docking behavior takes some experimenting to get a particular desired layout.

Ok... That's what I suspected. I just set the app's compatibility settings to this: 

image

and it works fine. It's the 4K monitor High DPI problem called out here https://github.com/CyberBoardPBEM/cbwindows/issues/19.

-Dale

GP/ToolPlay.cpp
DLLarson: Thanks for the great explanation! I wasn't aware MFC short circuited a lot of these messages.

-Dale

DLLarson:
>I suspect the problem across save/load is a duplicate of https://github.com/CyberBoardPBEM/cbwindows/issues/42.

Ok. We have it covered for future work.

-Dale

wsu-cb:
> Ok... That's what I suspected. I just set the app's compatibility settings to this:
> 
> image
> 
> and it works fine. It's the 4K monitor High DPI problem called out here #19 so it's not a new issue. I have to be more careful and change the app's properties before testing.
> 
> -Dale

I note that this image seems to say you set the compatibility settings to "Application", but #19's image seems to show the settings set to "System".  Could you clarify under what circumstances each is the better choice?

GShr/LBoxPiec.cpp
wsu-cb: > Just an FYI.... The Windows message `WM_COPYDATA` is used to pass data between processes.
> 
> -Dale

As I understand `WM_COPYDATA`, it can only be used to pass "simple" data, not pointers.  On the other hand, if the pointer refers to data that can be serialized, one could serialize the data to a memory buffer, and then pass the memory buffer.  But, even if we did that, I don't see much reason to pass pieces between different processes.

GM/Gm.cpp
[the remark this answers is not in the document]
wsu-cb: > I think your idea of adding the processID to the message name is a winner! It completely solves problem.

Actually, I had another idea later: instead of consuming a RegisterWindowMessage ID per process, move the current WPARAM drag phase information into the DragInfo structure pointed to by LPARAM, and then have WPARAM carry the process ID.  That would still allow the processes to protect themselves against using pointers from another process while sharing a single RegisterWindowMessage ID across all CB processes.  The limitation is that every out-of-process drag operation would have to return the same value, no matter what the drag phase is, but that seems OK to me.

DLLarson:
> I note that this image seems to say you set the compatibility settings to "Application", but #19's image seems to show the settings set to "System". Could you clarify under what circumstances each is the better choice?

Yes. That's kind of old. It only fixes the scaling of some windows but the docking drag/drop stuff is all out-of-whack. If I select Application the app is small and fuzzy but the coordinates are all working. 

I have to revisit this whole High DPI issue again and find a reliable solution. I haven't found definitive documentation on the DPI settings "dialog" so I really don't know the actual meaning of the choices between Application, System, and System (enhanced). Plus the Program DPI settings section of the dialog has an effect too.

-Dale

GShr/LBoxPiec.cpp
DLLarson: I only mentioned to point out that's the only sanctioned method using window messages. There are tons of ways to pass data between processes. DDEML (the predecessor to OLE/COM) for inter-application communication used messages and other tricks to move data between process.

We really have no reason to do this (well I could think of a few use cases for CBDesign but the clipboard already works fine for moving image data.)

-Dale

GM/Gm.cpp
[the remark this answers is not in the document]
wsu-cb: Dale, did you see the new idea I had in https://github.com/CyberBoardPBEM/cbwindows/pull/99#discussion_r1039039618?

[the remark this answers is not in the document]
DLLarson: >Dale, did you see the new idea I had in https://github.com/CyberBoardPBEM/cbwindows/pull/99#discussion_r1039039618?

No I missed it. Thanks for the note. 

This idea would work fine as well. I think both are solid, low risk, solutions to the problem. In the later solution it would also work should we wanted to send data from one process to another. 

-Dale

wsu-cb:
I have implemented https://github.com/CyberBoardPBEM/cbwindows/pull/99#discussion_r1039039618, so this is ready for another review.

wsu-cb:
> I've examined the code and did some testing with the 32 and 64 bit versions and it all works well. 

;-)

> Note that I've associated the drag/drop issue (#96) with this PR so it'll be closed when we merge this pull request.

Commit d29af already says it will fix #96.  Did you overlook that, or does this association mean something different?

> I'm good with merging this to master with your thumbs up.

Yes, I think this is ready.

DLLarson:
>Did you overlook that, or does this association mean something different?

I didn't see that at all. In fact I couldn't find a way to associate a PR with an issue--just a way to associate an issue with a PR. The difference I had to do that in the issue's interface. Anyway, it's not a big deal either way.

Merging to master.....

-Dale

PR #101, opened 2022-12-11
implement CTilePalette::OnGetDragSize() to fix DRAG_TILE

### LINKED ISSUE #96: Handle Drag/Drop of selected objects into incompatible target views

wsu-cb:
Fix #96


DLLarson (approved):
That fixes it!

I'll go ahead and merge it.

-Dale

PR #102, opened 2022-12-13
Serialize crollstate

wsu-cb:


DLLarson (commented):
The die roller code seems to be working. To check it made some moves and played them back. However, when I played them back I ran into the issues shown here:

image
****
These may just need a new issue but I thought I'd point them out here and let you decide.

-Dale

wsu-cb:
> These may just need a new issue but I thought I'd point them out here and let you decide.

Could you please delete the CB registry and re-test?  I have seen that changing between different branches/tags sometimes causes resource ID problems with stuff cached in the registry.

DLLarson:
> > These may just need a new issue but I thought I'd point them out here and let you decide.
> 
> Could you please delete the CB registry and re-test? I have seen that changing between different branches/tags sometimes causes resource ID problems with stuff cached in the registry.

Good call! That fixed the toolbar buttons. Now I can step using the toolbar buttons.

However, the Done button issue still exists. I'll add a new issue for that.

-Dale

DLLarson (approved):
The die roller serialization looks good. It's nice not to have to remember what you were rolling between sessions! 

Let me know if I should proceed to merge this PR.

-Dale

wsu-cb:
> Let me know if I should proceed to merge this PR.

I think this should be merged.  I'm reasonably sure that the button text and object ID problems we are seeing are unrelated to this.

wsu-cb:
> > Let me know if I should proceed to merge this PR.
> 
> I think this should be merged. I'm reasonably sure that the button text and object ID problems we are seeing are unrelated to this.

No, wait.  I think this SHOULD NOT be merged.  I think the object ID problem involves a file-format conversion operation.  As such, I think that should be done in this MR to avoid another MR with file conversion changes.

DLLarson:
>I think this SHOULD NOT be merged.

Ok... No problem. 

-Dale

DLLarson (approved):
It's loading the older games again. I like the fix. Nice and simple.

I think it's ready to merge to master on your command.

-Dale

wsu-cb:
> I think it's ready to merge to master on your command.

Yes, I think this is ready to merge now.

PR #104, opened 2022-12-17
fix button text

### LINKED ISSUE #103: Game History Playback Entry Has "Done" button when it should be "Replay" and "Export".
When history entry is activated for playback the associated button shows "Done" when I should show "Replay" and "Export" buttons. If the history entry is single clicked it fixes the buttons. If the history entry is double clicked, the "Done" button is shown again.

image

-Dale

wsu-cb:
Programmatic SetCurSel() doesn't trigger lbox window handler

Fix #103


DLLarson (approved):
Perfect! Since it's so straight forward I'll merge right away.

-Dale

PR #111, opened 2023-01-02
CTileManager::FindNamedTileSet():  fix reversed test

### LINKED ISSUE #110: CTileManager::FindNamedTileSet() reversed sense?
https://learn.microsoft.com/en-us/cpp/atl-mfc-shared/reference/cstringt-class?view=msvc-170#comparenocase says that the strings are equal when the `CompareNoCase()` return value is `0`.  Given that, it looks like `CTileManager::FindNamedTileSet()` is using `CompareNoCase()` backwards.  I have confirmed that tag `v3.5-prerelease-2` uses the same logic as the current version, but it still looks wrong to me.  Dale, can you confirm that `CTileManager::FindNamedTileSet()` is currently buggy?

wsu-cb:
Fix #110

### DIFF
--- a/GShr/TileMgr.cpp
+++ b/GShr/TileMgr.cpp
@@ -214,7 +214,7 @@ size_t CTileManager::FindNamedTileSet(const char* pszName) const
     {
         const CTileSet& pTSet = GetTileSet(i);
         CString strName = pTSet.GetName();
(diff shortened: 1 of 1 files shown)


DLLarson (approved):
Looks good!
-Dale

PR #112, opened 2023-02-15
Vs2022

wsu-cb:
I have included an "extra" commit at the end of this PR that demonstrates some of the C++20 `std::format` capabilities.  Since VS2022 Community Edition already provides `std::format` (VS2022 is where I developed this demo), I hope we can make using C++20 a requirement.  The trouble with using `sprintf` is both that it's not type safe (and you know how I feel about type safety), and, even worse, MS goes against the C++ standard for interpreting `%s` in Unicode (https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170), which makes it painful to write portable code along with current `wxWidgets` (https://github.com/wxWidgets/wxWidgets/commit/fec8c06a53f5f2eb65e3328eadddd22334e7971f).

The "extra" commit should be removed before applying this PR, so I am marking this PR as draft currently.


DLLarson:
>I hope we can make using C++20 a requirement. 

Is this level of the standard available in current *nix environments via GCC or CLANG?

If so, it's a no brainer.

-Dale

DLLarson (commented):
I've looked over the code changes and the std::format demos. I'm good with all of it so long as targets for *nix support the c++20 stuff.

wsu-cb:
> I've looked over the code changes and the std::format demos. I'm good with all of it so long as *nix also supports the C++20 stuff for wxWidgets.
> 

And that's the potential gotcha currently.  Clang 14 (March 2022) is supposed to have `std::format`, but it must be explicitly turned on.  GCC doesn't have `std::format` yet.  It's supposed to be in GCC 13, and it looks likely that will be available by mid-2023, but there's no guarantee.  On the other hand, I figure by the time I get CB ported to wxWidgets for MSVC, `std::format` may be available on Linux.  If it's not, there's also https://fmt.dev/latest/api.html#compatibility-with-c-20-std-format, which I think actually provided a lot of the initial examples for `std::format`.

wsu-cb:
> > I hope we can make using C++20 a requirement.
> 
> Is this level of the standard available in current *nix environments via GCC or CLANG?
> 
> If so, it's a no brainer.
> 
> -Dale

According to https://en.cppreference.com/w/cpp/compiler_support/20, Linux support for most of C++20 is available now, except for Modules and https://github.com/CyberBoardPBEM/cbwindows/pull/112#issuecomment-1432461816.  However, some of C++20 requires recent compiler versions.

For me, `std::format` is the only strong reason to require anything newer than C++11, so I could see staying with C++11 on Linux and using `fmtlib` (https://github.com/fmtlib/fmt) instead.  We could even use `fmtlib` on Windows instead of `std::format`, but that would require using `CMake` or` vcpkg`, so I would prefer to stick with MSVC C++20 and postpone dealing with `fmtlib` until we see (lack of) `std::format` becoming a serious problem.

DLLarson:
>so I would prefer to stick with MSVC C++20 and postpone dealing with fmtlib until we see (lack of) std::format becoming a serious problem.

Yep. Let's just forge ahead. It's not like we're switching to 'nix in the near future but I would imagine when the code gets feature stable we could try switching then. I'm not sure how to approach that port. Last time we started from scratch.

One hassle I've hit with the DPI awareness work is that the owner draw list boxes are limited to 255 pixels in height regardless of screen scaling. That totally sucks. That cap goes all the way back to Win16. However, wxWidgets has a virtual listbox class that doesn't appear to have that nasty limitation.

-Dale

wsu-cb:
> > so I would prefer to stick with MSVC C++20 and postpone dealing with fmtlib until we see (lack of) std::format becoming a serious problem.
> 
> Yep. Let's just forge ahead. It's not like we're switching to 'nix in the near future but I would imagine when the code gets feature stable we could try switching then. I'm not sure how to approach that port. Last time we started from scratch.
> 

My current thinking is:

1)  Get to VS2022
2)  Create a class `CB::string` that wraps a `std::string` and also provides the same quick-and-dirty ANSI/Unicode conversion that I use in the temporary commit's demo to allow convenient conversion to a Windows UCS2 string
3)  Convert `sprintf`, etc to `std::format` using `CB::string`
4)  Use `CB::string` to get a mostly-working Unicode build of CB
5)  The wxWidgets suggest it can co-exist with MFC, so try to splice it into 4.  The reason for the previous from-scratch port was that I didn't have the extra GUI toolkit that CB3 used.
6)  Try to replace a single MFC dialog with a wxWidgets dialog as a proof-of-concept.
7)  Assuming 6 works, just keep going piece-by-piece


> One hassle I've hit with the DPI awareness work is that the owner draw list boxes are limited to 255 pixels in height regardless of screen scaling. That totally sucks. That cap goes all the way back to Win16. However, wxWidgets has a virtual listbox class that doesn't appear to have that nasty limitation.
> 
> -Dale

I have been aware of that limit because I have an unpublished `.gbx` that has card images where the cards are too tall to display in a Tray.  It would be nice to have that fixed, but that's a long way in the future.

DLLarson:
>The wxWidgets suggest it can co-exist with MFC, so try to splice it into 4. The reason for the previous from-scratch port was that I didn't have the extra GUI toolkit that CB3 used.

This sounds promising but I don't fully understand how this co-existence is provided. Do you have a link?

-Dale

DLLarson:
Looks good. I'm ready to merge to `master` if you're good with it.

-Dale

wsu-cb:
> > My current thinking is: ..... The wxWidgets suggest it can co-exist with MFC, so try to splice it into 4. The reason for the previous from-scratch port was that I didn't have the extra GUI toolkit that CB3 used.
> 
> This sounds promising but I don't fully understand how this co-existence is provided. Do you have a link?
> 
> -Dale

https://github.com/wxWidgets/wxWidgets/blob/master/samples/mfc/mfctest.cpp

wsu-cb:
> Looks good. I'm ready to merge to `master` if you're good with it.
> 
> -Dale

Yes, I'm ready to merge.

PR #114, opened 2023-02-19
update .sln file to VS2022

wsu-cb:
I have no idea why I now finally see the `.sln` file updated for VS2022, when I didn't see it updated before, but here it is...


DLLarson (approved):
Makes sense. VS2022 is internally version 17 of the toolchain. I learned this by building wxWidgets recently. I don't think solution files really change very much. The older one worked just fine.

-Dale

PR #125, opened 2024-02-04
Xrc part02

wsu-cb:
Here are some more conversions of CBDesign dialogs from MFC to wx.

Note that these include some uses of `wxVListBox`.  In keeping with approximately the same layout as the original dialogs, none of the current replacements has enough height to demonstrate the behavior of `wxVListBox` with tiles of height > 255 pixels.  However, I did experiment with changing the height of the Create Marker dialog enough to verify that `wxVListBox` does correctly display tiles 350 pixels high when the `wxVListBox` is tall enough to show the entire tile.  However, the scrolling behavior of `wxVListBox` is not ideal when a tile exceeds the height of the window:  `wxVListBox` does not support scrolling of partial lines; the top of the topmost visible item must always be at the top of the `wxVListBox`, so it is still impossible to see the bottom of a tile if the tile exceeds the `wxVListBox` height.


wsu-cb:
Closed/reopened PR to force checks to run again because, for reasons I don't understand, the checks were interrupted.

GShr/LBoxMark.cpp
```diff
@@ -254,3 +260,197 @@ BOOL CMarkListBox::OnDragSetup(DragInfo& pDI) const
...
+{
+    if (m_eTrayViz == mtrayVizNormal)
+    {
+        if (m_pDoc == NULL)
+            return;
+
+        CMarkManager* pMMgr = m_pDoc->GetMarkManager();
+        if (pMMgr == NULL)
+        {
+TRACE("handled null pMMgr\n");
```
DLLarson, line 416: Indent TRACE.

DLLarson:
>However, the scrolling behavior of wxVListBox is not ideal when a tile exceeds the height of the window: wxVListBox does not support scrolling of partial lines; the top of the topmost visible item must always be at the top of the wxVListBox, so it is still impossible to see the bottom of a tile if the tile exceeds the wxVListBox height.

What did the Windows do when it couldn't show the entire item of the 250-ish height limit?

-Dale

DLLarson (approved):
I'm done. Not much to comment on. Except for the wxWidgets virtual list box most is just one-for-one switch overs from MFC to wxWidgets.

Which method of merge would you prefer?

-Dale

GShr/LBoxGrfx.h
```diff
@@ -428,5 +429,352 @@ class CGrafixListBoxData : public BASE_WND
...
+class CGrafixListBoxWx : public CB::VListBoxHScroll
+{
+    // Construction
+public:
+    CGrafixListBoxWx();
+    ~CGrafixListBoxWx() override;
+
+    // Attributes
+public:
+#if 0
```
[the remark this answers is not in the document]
wsu-cb: If I left any in that aren't needed, that was a mistake.  However, I have definitely used `#if 0` for things that aren't needed YET, but which I expect will be needed eventually.

wsu-cb:
> > However, the scrolling behavior of wxVListBox is not ideal when a tile exceeds the height of the window: wxVListBox does not support scrolling of partial lines; the top of the topmost visible item must always be at the top of the wxVListBox, so it is still impossible to see the bottom of a tile if the tile exceeds the wxVListBox height.
> 
> What did the Windows do when it couldn't show the entire item of the 250-ish height limit?
> 

I didn't remember, so I had to go recompile that version.  It turns out the MFC version behaves the same way.

GShr/LBoxMark.cpp
wsu-cb: Actually, if I do anything, I will just remove this line.  Do you want me to do that before merging the PR?

wsu-cb:
> I'm done. Not much to comment on. Except for the wxWidgets virtual list box most is just one-for-one switch overs from MFC to wxWidgets.
> 
> Which method of merge would you prefer?
> 

I prefer keeping the individual commits, but if you prefer to squash them, it won't bother me very much.

DLLarson:
>I prefer keeping the individual commits, but if you prefer to squash them, it won't bother me very much.

Your wish is my command...after the final push is done that is.

GShr/LBoxMark.cpp
wsu-cb: done

PR #132, opened 2024-07-07
convert Bit Editor (and associated selector) from MFC to wxWidgets

wsu-cb:
The MFC doc/view framework is not being replaced yet, but, other than that, convert as much as is feasible of VwTilesl and VwBitedt from MFC to wxWidgets .


DLLarson (approved):
Finished the review and exercised the application. All seems to be working very well.

Two things though.

- I think the "apply board mask" feature should allow undo. I know it doesn't now but it should be easy since bitmap editing already has support. I think this was an oversight on my part.
- The docking of the color palette draws tracking rectangles that are way out of place. Do you see the same on yours without display scaling like I get for having a 4k screen?

Dale

wsu-cb:
> Two things though.
> 
> * I think the "apply board mask" feature should allow undo. I know it doesn't now but it should be easy since bitmap editing already has support. I think this was an oversight on my part.

I will look into this.  Should I add that to this PR, or make it a separate PR?  I see you already approved this PR, but you haven't merged it yet, so I am inclined to append it here rather than create a new PR.

> * The docking of the color palette draws tracking rectangles that are way out of place. Do you see the same on yours? I have display scaling on for having a 4k screen while I don't think you do.

I don't see anything strange with moving the color palette.  You are correct that I am using a 15" laptop screen.  The docking window code has not been converted from MFC to wx.  Can you test whether the `v3.5-prerelease-2` tag behaves the same way?  Also, does the `Tiles` palette have the same problem?

> 
> Dale

DLLarson:
>I will look into this. Should I add that to this PR, or make it a separate PR? I see you already approved this PR, but you haven't merged it yet, so I am inclined to append it here rather than create a new PR.

Just add it to this PR. That shouldn't be an issue. In fact, my guess is when some additional changes are pushed it should go into inspection mode. Let's find out.

DLLarson:
>I don't see anything strange with moving the color palette. You are correct that I am using a 15" laptop screen. The docking window code has not been converted from MFC to wx. Can you test whether the v3.5-prerelease-2 tag behaves the same way? Also, does the Tiles palette have the same problem?

The docking seems to working properly when I run `v3.5-prerelease-2`. There are other problems due to the HighDPI compatibility but the docking code works for both the color and tile palettes. In the current code dragging the Color palette shows this:

image

Dragging the Tile palette is the same:

image

-Dale

wsu-cb:
> If yours looks fine, could you try a brief experiment and set the system/display settings from 100% scaling to 125% and try dragging to confirm this is a High DPI scaling problem. wxWidgets supposedly should work in HighDPI settings so there might be some strange interactions with the mixed setup.
> 

Your hypothesis that setting 125% scaling would cause the palette tracking rectangles in the new code to go haywire, but that v3.5-prerelease-2 would not have a problem, is correct.

However, after doing some more investigation, I have found that the problem isn't exactly wxWidgets:  If I compile using the `.sln` instead of CMake, the tracking rectangles work.  Also, if I change the `.sln` to state that the application is NOT DPI aware then the tracking rectangles go haywire.  Therefore, the problem is related to the application manifest.  So the problem is how to configure CMake and wxWidgets together to get the application manifest to report DPI awareness.

(I am out of time to work on this for now, but I thought you would find this interesting.)

wsu-cb:
I have found a way to get DPI awareness enabled.

DLLarson:
Hi Bill,

Well I've played around with results and the docking does indeed seem to be working. However, since the app is assumed to be HiDPI aware, nothing is scaled to the systems settings. The color palettes is kinda crowded and the board graphics are--on my 150% scaled 4k 32" screen--small. When DPI awareness is on the Windows system renders the app of screen, then scales it to the target device. 

Another side effect, suddenly the application icon for CBDesign.exe has changed to something else:

image

I like to idea of making the application ACTUALLY DPI aware. WxWidgets supplies apis for DPI based rendering. It seems like that's the way to go rather count on Windows to scale in compatibility mode and produce a fuzzy looking application that breaks the docking code of the framework.

-Dale

DLLarson:
Hi Bill,

I got ambitious today and took a whack at making the color palette DPI compliant. I think I succeeded. The wxWidgets DPI support was pretty simple for this case anyway. The tile/piece style lists may not be too bad either.

The main board windows have a lot of scaling trickiness going on so that still going to be the toughest nut to crack.

Could you please apply the attached patch file and give it a look?

[0001-Add-wxWidgets-DPI-awareness-to-the-color-pallete-too.patch](https://github.com/user-attachments/files/16317376/0001-Add-wxWidgets-DPI-awareness-to-the-color-pallete-too.patch)

-Dale

wsu-cb:
> Another side effect, suddenly the application icon for CBDesign.exe has changed to something else:
> 

I overlooked this.  I think I have a work-around, but it's a little grubby.  Also, Windows apparently caches those icons in some fashion.  I was worried that my work-around didn't work, but after rebooting, the icon change took effect.

wsu-cb:
> Hi Bill,
> 
> I got ambitious today and took a whack at making the color palette DPI compliant. I think I succeeded. The wxWidgets DPI support was pretty simple for this case anyway. The tile/piece style lists may not be too bad either.
> 
> The main board windows have a lot of scaling trickiness going on so that still going to be the toughest nut to crack.
> 
> Could you please apply the attached patch file and give it a look?
> 
> [0001-Add-wxWidgets-DPI-awareness-to-the-color-pallete-tool-v2.patch](https://github.com/user-attachments/files/16321512/0001-Add-wxWidgets-DPI-awareness-to-the-color-pallete-tool-v2.patch)
> 
> -Dale

I have run with this patch, and don't see any obvious problems.  It makes the Color Palette pretty big on my screen when running at 125%, but presumably that's the point.

However, I have a couple of concerns with the patch as written:
1.  The `CColorPalette` member functions are using the `CDockColorPalette`'s `FromDIP()`.  Why doesn't `CColorPalette` use its own `FromDIP()`?  It seems pretty unlikely that they would return different results, but presumably it could happen in theory.
2.  `CColorPalette::ComputeLayout()` is actually getting called during `CMainFrame::OnCreate()` before `CColorPalette::Create()` is called, and thus `CColorPalette::GetParent()` returns `nullptr`.  `CColorPalette::Create()` then calls `FromDIP()` with that `nullptr`.  We are getting away with it currently, but that's Undefined Behavior.  It's not clear to me how to solve this problem; it looks like the intent is for `CMainFrame::OnCreate()` to make layout decisions before creating its children, but I don't know if we can assume `FromDIP()` is valid for an uncreated window.  I have posted that question to the wxWidgets forum (https://groups.google.com/g/wx-dev/c/uHeHzkIsO3Q/m/aTJh8UW8AAAJ).  We might have to just accept that the initial call to `CColorPalette::ComputeLayout()` should use `CMainFrame::FromDIP()`.

DLLarson:
Hi Bill,

>I have run with this patch, and don't see any obvious problems. It makes the Color Palette pretty big on my screen when running at 125%, but presumably that's the point.

Completely the point. Below I have some screenshots on my 4k monitor of both shipping and this development code. You can see the color palette tool sizes now track each other.

> However, I have a couple of concerns with the patch as written:
> 
> 1. The `CColorPalette` member functions are using the `CDockColorPalette`'s `FromDIP()`.  Why doesn't `CColorPalette` use its own `FromDIP()`?  It seems pretty unlikely that they would return different results, but presumably it could happen in theory.

I just didn't think the `CColorPalette` had wxWindow ancestry. This makes the code much cleaner. 

> 2. `CColorPalette::ComputeLayout()` is actually getting called during `CMainFrame::OnCreate()` before `CColorPalette::Create()` is called, and thus `CColorPalette::GetParent()` returns `nullptr`.  `CColorPalette::Create()` then calls `FromDIP()` with that `nullptr`.  We are getting away with it currently, but that's Undefined Behavior.  It's not clear to me how to solve this problem; it looks like the intent is for `CMainFrame::OnCreate()` to make layout decisions before creating its children, but I don't know if we can assume `FromDIP()` is valid for an uncreated window.  I have posted that question to the wxWidgets forum (https://groups.google.com/g/wx-dev/c/uHeHzkIsO3Q/m/aTJh8UW8AAAJ).  We might have to just accept that the initial call to `CColorPalette::ComputeLayout()` should use `CMainFrame::FromDIP()`.

The code actually defaults to data from the WindowDC.

However, looking at all that rigmarole I find that it's really not needed. 

The sizes of windows are all automatically recomputed once the window hierarchy is in place. So I removed all that stuff and it runs just fine.

So you can dispense with my last patch and try this:

[0001-Add-wxWidgets-DPI-awareness-to-the-color-pallete-tool-v3.patch](https://github.com/user-attachments/files/16322741/0001-Add-wxWidgets-DPI-awareness-to-the-color-pallete-tool-v3.patch)

An aside....

Now for a pleasant surprise I got when I was testing against a gamebox that just happened to have playing cards that are taller than the 254 pixels Windows owner draw listboxes could handle so they just got chopped off.

So here is the current shipping CBDesign:

image

Note the truncated playing cards. Now look at a screenshot of the code I just built with my latest patch (above):

image

You can see the entire cards!!!! That is so sweet! You did a fantastic job converting the graphical list boxes. Thanks for that!

-Dale

wsu-cb:
I have some suggested amendments to your patch.  Please see 
[0001-CColorPalette-DPI-awareness-amendments.patch](https://github.com/user-attachments/files/16323098/0001-CColorPalette-DPI-awareness-amendments.patch)

DLLarson:
> I have some suggested amendments to your patch. Please see...

Good catches! (I thought I removed all the dead code--this is why we do code inspections.)

Please apply to the PR with the others if you're happy with the additions.

-Dale

DLLarson (approved):
All looks good to me. I'll merge the PR if you feel it is complete.

I did get one error when doing a test build and run. It happened when I was setting colors in the color palette but I couldn't reproduce it.

image

-Dale

wsu-cb:
> All looks good to me. I'll merge the PR if you feel it is complete.

Yes, merge it please.

> I did get one error when doing a test build and run. It happened when I was setting colors in the color palette but I couldn't reproduce it.
> 
> I let it continue and things seemed fine. I just thought I'd make a note of it just-in-case it's meaningful to you.
> 

I'm pretty sure I have seen that assert a couple of times, but I haven't been able to reproduce it for investigation.

DLLarson:
>Yes, merge it please.

Done.

PR #134, opened 2024-11-10
convert VwPrjgbx to wxWidgets

wsu-cb:
Note that, as of this writing, the current `.fbp` file is not readable by the master branch of wxFormBuilder.  I have started a PR to add `wxVListBox` support to wxFormBuilder (https://github.com/wxFormBuilder/wxFormBuilder/pull/858), but it's still a work in progress.


DLLarson:
Just saw this. (For some reason I didn't see a notice when it was posted.)

-Dale

DLLarson:
Hi Bill,

Is this a work in progress?

I ask because I see large blocks of `#if 0` code removals in some files.
Or are these to be removed later?

-Dale

wsu-cb:
> Hi Bill,
> 
> Is this a work in progress?
> 
> I ask because I see large blocks of `#if 0` code removals in some files. Or are these to be removed later?
> 
> -Dale

This is ready for review.  However, you are correct that there is a significant amount of `#if 0` code. Some of it should be removed after I/we are confident that there aren't any problems with the new code where having the old code readily accessible would be convenient.  (Mostly, the VwPrjgbx `#if 0` code.)  There is other code (mostly the LBoxGrfx code) where the code is commented out because it's not needed yet, so I don't want to bother trying to port it yet.  Basically, I currently think that, when I expect to have both the MFC and wx code existing simultaneously, it's better to copy all the MFC code and then start converting to wx as needed.

DLLarson:
>I currently think that, when I expect to have both the MFC and wx code existing simultaneously, it's better to copy all the MFC code and then start converting to wx as needed.

I totally agree. As they say... Eat that elephant a bite at a time.

Ok... I'll finish it up soon. It's not as big as some of the previous PR's.

-Dale

DLLarson (approved):
Done. Nothing more to say about this PR. It runs fine with various gameboxes I threw at it. Most changes are 1-to-1 mods to object indirection and use of wx* vs MFC's equivalents.

-Dale

DLLarson:
It looks like we shouldn't squash on merge.

Waiting for your thumb's up.

-Dale

DLLarson:
Whoops... I guess I missed your literal thumbs up! 

I guess I don't get those in email notices.

Done.

-Dale

PR #135, opened 2024-12-12
convert GM CGamDoc from MFC to wx

wsu-cb:


DLLarson:
Hi Bill,

Just a heads up... I see your PR but won't be able to get to it for at least a couple days.

-Dale

DLLarson:
Sorry, this really got away from me over the holidays. I'm on it now.
-Dale

DLLarson (approved):
Finished looking at the code and all looks fine to me. Many mods are just redirection changes and most of the rest are more boilerplate in nature as we go.

During some cursory run tests I noticed that the `CBPlay` program is scaling the view and scrollbox bitmaps and `CBDesign` is not. I think this is because you haven't really started the conversion of the `CBPlay` codebase yet. I just mention it in case I'm wrong. 

One other strange thing is when I first select a GAM or GSN file it will tell me it's going to upgrade the older format (which is totally fine) but then says it can't find the GBX file yet the message box shows the path and the file is definitely at that path. The second time through it found it fine even though I didn't save any files. I think it may have something to do with the program's starting "current working directory".

I'll merge if you're happy with this assessment. Sorry about the review delay. We had a hectic holiday season this year.

-Dale

wsu-cb:
> During some cursory run tests I noticed that the `CBPlay` program is scaling the view and scrollbox bitmaps and `CBDesign` is not. I think this is because you haven't really started the conversion of the `CBPlay` codebase yet. I just mention it in case I'm wrong.
> 

I have changed the resource and manifest configuration for `CBDesign` but not for `CBPlay`, so I would expect them to behave differently.  I hope the `CBDesign` behavior is better than the `CBPlay` behavior, but your description here doesn't make it clear to me which is better.  From the previous MR discussions, I assume `CBDesign` is behaving correctly, so I will leave this alone for now.

> One other strange thing is when I first select a GAM or GSN file it will tell me it's going to upgrade the older format (which is totally fine) but then says it can't find the GBX file yet the message box shows the path and the file is definitely at that path. The second time through it found it fine even though I didn't save any files. I think it may have something to do with the program's starting "current working directory".
> 

I don't remember seeing this, but if it only happens the first time one runs, I would expect not to remember it.  For now, I won't do anything about this either.

> I'll merge if you're happy with this assessment. 
> 

Yes, please merge.

DLLarson:
> but your description here doesn't make it clear to me which is better. From the previous MR discussions, I assume CBDesign is behaving correctly, so I will leave this alone for now.

I like the rescaled version of `CBPlay` best because they are a good size for my 4k HiDPI scaling. Otherwise the board and piece details are way too small.

HOWEVER.... I prefer the `CBDesign` behavior best at this time because I think it's better that the program rescale the views rather than get the fuzzy rescaling that Windows does behind the scenes. 

So this is a can kicked down the road for now until we get enough of wxWidgets capabilities are available to handle the HiDPI scaling natively.

-Dale

DLLarson:
Merged.

-Dale

PR #138, opened 2025-08-16
Gp dlgs+gsn

wsu-cb:


DLLarson (approved):
This is the smallest PR in a long while. 

I see you're onto the CBPlay code now which is really great. I have nothing to call out in this PR. 

However, the resource change that indicates that the app handles HighDPI processing should be changed to match CBDesign at some point sooner later to see what will break. I'll merge when you give the thumbs up.

Thanks!
-Dale

wsu-cb:
> This is the smallest PR in a long while.
> 
> I see you're onto the CBPlay code now which is really great. I have nothing to call out in this PR.
> 
> However, the resource change that indicates that the app handles HighDPI processing should be changed to match CBDesign at some point sooner later to see what will break. I'll merge when you give the thumbs up.
> 
> Thanks! -Dale

I just noticed that this hasn't been merged.

DLLarson:
>I'll merge when you give the thumbs up.

I'll take that as a "Thumbs up"? I usually want you're final say after the review.

-Dale

PR #139, opened 2026-03-21
convert CPlayBoardView to wx

wsu-cb:
I haven't given up working on CyberBoard; it just took a lot of work to convert `CPlayBoardView` to wxWidgets.  Unfortunately, correspondingly, this is a very large Pull Request.  One reason this is so large is that, after converting `CPlayBoardView`, I found that it was impossible to play a game because I couldn't use the MFC Marker and Piece palettes with the wx `CPlayBoardView`, so I also converted those palettes in this PR.


wsu-cb:
I am still using MSVS 2022, and I suspect your recent download is MSVS 2026.  I updated the CMake version specification to match what wxWidgets is using.  Please see if that solves your problem.

DLLarson:
That fixed both forms of command line builds. Thank you!

You'll enjoy this...Here's what I get when I run the app:

image

Following the "Learn More" link:

https://support.microsoft.com/en-us/windows/smart-app-control-frequently-asked-questions-285ea03d-fa88-4d56-882e-6698afdb7003#bkmk_unknown

```
When you try to run an app on Windows, Smart App Control will check to see if our 
intelligent cloud-powered security service can make a confident prediction about 
its safety. If the service believes the app to be safe, Smart App Control will let 
it run. If the app is believed to be malicious or potentially unwanted, then 
Smart App Control will block it.

If the security service is unable to make a confident prediction about the app, then 
Smart App Control checks to see if the app has a valid signature. If the app has 
a valid signature, Smart App Control will let it run. If the app is unsigned, or 
the signature is invalid, Smart App Control will consider it untrusted and block 
it for your protection.
```

Remember I said I did a fresh install of Windows? Well this is why that matters:

```
Smart App Control is only available on clean installs of Windows 11. Also, there 
are other reasons why Smart App Control could be turned off.
```

I fear we'll get this thing ready and M$ will block it for folks who have nice shiny new installs. They're trying to force code signing down people's throats.

I turned the "feature" off so I can do some work. The way I read it, it should never allow itself to be turned back on again but a registry tweak can turn on.

The weird thing is it allowed the CBPlay app to run without issue but CBDesign is rejected and not allowed to run. Both were debug builds.

Oh, well... Forward we go.

wsu-cb:
> That fixed both forms of command line builds. Thank you!

Good to hear!

> ```
> When you try to run an app on Windows, Smart App Control will check to see if our 
> intelligent cloud-powered security service can make a confident prediction about 
> its safety. If the service believes the app to be safe, Smart App Control will let 
> it run. If the app is believed to be malicious or potentially unwanted, then 
> Smart App Control will block it.
> 

Yikes.  I haven't run into that, but my machine is ~5 years old.

> I fear we'll get this thing ready and M$ will block it for folks who have nice shiny new installs. They're trying to force code signing down people's throats.
> 

I'm not thrilled, but when even Notepad++ gets hijacked, I can understand the attraction of "security theatre".

> The weird thing is it allowed the CBPlay app to run without issue but CBDesign is rejected and not allowed to run. Both were debug builds.

Bizarre.

> 
> Oh, well... Forward we go.

And code signing certificates have recently been limited to validity periods of ~15 months, so in practice they need to be updated annually.  Nevertheless, it looks like we will have to go down that path eventually if CB is to continue as a viable project.

DLLarson (changes requested):
This is a large set of changes. Many are access changes (ref vs ptr vs...) but there is a lot of logic mapped to the wx equivalent calls. I hope this is an accurate assessment of what I see.

I gave the code a cursory review. It's pretty large and as such, very tough to detect logic issues and the like. I did put in a bunch of notes to make the Copyrights 2026. I think if a module has a code change it should update the (C). I probably missed a bunch of these since the deltas may not show the copyright. I'm still need to run the code with various text gameboxes, etc. for a sanity check.

DLLarson:
>The autoscrolling is now almost entirely managed by wxWidgets, not CB. (I actually made some wx submissions to make the autoscrolling suitable, but what wx originally provided was good enough that I thought improving it was better than converting the CB-specific autoscroll code.)

Very cool! The CB code actually started life in Win16 (Borland C++ w/OWL) days so lots of stuff had to be hand rolled then to just work or for other performance issues. 

For example all the specialized font redundancy handling was because Win16 was very slow realizing fonts in the early days plus you couldn't have tons of HFONT's laying around. So it maintained a single font realization (logical font -> rendering font) per font instance. It was modeled on Windows atom strings where a single string was registered as an atom and an atom's handle was passed around for speed. DDE is full of those. That wouldn't be an issue in current times. Win16 memory was cramped plus CPU's were so much slower.

>Or are you saying the merge date, not the commit date, is what should be represented by the copyright?

Yes. I'm looking at it from the perspective of when it was merged. I think this is when it was published. I believe the law says it was copyrighted upon creation. But my way of thinking it is that a book writer doesn't copyright the text as each page is written or edited but when it was made available for outside consumption.

It's not a huge deal either way but just something that I noticed. I pretty much figured that you updated them at the moment you edited them and time passed into the new year.

Regardless, I'll merge when you give a thumbs up and I give the code some run time.

Thanks for all your work!

-Dale

DLLarson:
Hi Bill,

I've run both CBDesign and CBPlay (which still has the old look with views that are  properly scaled) and both run nicely. The only odd artifact I observed is double repaints of the board view in CBDesign. This is very visible with complex boards like PoG with lots of rendered graphical elements. The board is drawn, then erased, and re-rendered.

An easy way to see this to tab away to the project page and then reselect the board tab. 

The palettes operate very cleanly and look great. In fact the whole CBDesign program looks very crisp. The only issues (other the double renders mentioned above) are what we already know--the HighDPI view and tiles stuff. 

I also did discover that if I change the display scaling after running CBDesign and rerun the program the UI sizes of some stuff is wonky. I have to delete the registry settings to get it to look right. So some display scaling information is being stored in the program's registry.

I'm going to merge.

Thanks again for the great work. CBDesign is really starting to feel polished!

-Dale

DLLarson (approved):
Ready to rock...

wsu-cb:
> The only odd artifact I observed is double repaints of the board view in CBDesign. This is very visible with complex boards like PoG with lots of rendered graphical elements. The board is drawn, then erased, and re-rendered.
> 
> An easy way to see this to tab away to the project page and then reselect the board tab.

I have reproduced this, and have found a workaround.  I will submit it shortly.

> Thanks again for the great work. CBDesign is really starting to feel polished!

I did fix a couple CBDesign bugs in this PR, but I want to double-check that you understood that this PR mainly affects CBPlay.

DLLarson:
>I did fix a couple CBDesign bugs in this PR, but I want to double-check that you understood that this PR mainly affects CBPlay.

Actually I did miss that part. Thanks for pointing that out. I did exercise the CBPlay program as well--including stepping through some move histories--but saw the whole PR as more spread out file-wise once I started looking through it. No matter. 

As bonus we discovered and corrected some issues with using the current Visual Studio Community drop.

PR #140, opened 2026-03-31
reduce unnecessary repaint

wsu-cb:
wxAUI invalidates windows that changed size when the idle processing shows/hides palettes, so defer painting a newly-activated tab until idle processing has finished with the palettes


DLLarson (approved):
The double paints are now gone and the overall effect is much better. 

I do wonder what is causing the long render times before the board image is displayer. I see a blank area, long delay, then the final full render.

My system is pretty beefy for graphics GPU (RTX-4080 Super) and CPU (i9-14gen) so this shouldn't be happening. I think we must still have some bitmap conversions for somethings going on behind the scenes. 

Just thought I'd make note of it.

Gonna merge...

-Dale

